PR #10, opened 2026-01-29
feat: add websocket setup

xsachax:
Add ws server, communication protocol (event types) and connection hub as well as a test client. Basic implementation of setup in main, but no events being produced yet. TODO: forward commands to debugger once that part is implemented

### DIFF
--- a/internal/ws/ws_test.go
+++ b/internal/ws/ws_test.go
@@ -23,42 +23,57 @@ func TestWebSocket(t *testing.T) {
 }
 
 var _ = Describe("Hub", func() {
+	var (
+		hub            *Hub
+		server         *httptest.Server
+		wsURL          string
+		shutdownCalled *atomic.Bool
+	)
+
+	BeforeEach(func() {
+		hub = NewHub("test-session", time.Minute)
+		shutdownCalled = &atomic.Bool{}
+		hub.onShutdown = func(sessionID string) {
+			shutdownCalled.Store(true)
+		}
+
+		go hub.Run()
+
+		server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			upgrader := websocket.Upgrader{}
+			conn, _ := upgrader.Upgrade(w, r, nil)
+			defer func() { _ = conn.Close() }()
+		}))
+
+		wsURL = "ws" + strings.TrimPrefix(server.URL, "http")
+	})
+
+	AfterEach(func() {
+		if server != nil {
+			server.Close()
+		}
+	})
+
 	Describe("NewHub", func() {
 		It("should create a new hub with correct properties", func() {
 			sessionID := "test-session"
 			idleTimeout := 5 * time.Minute
 
-			hub := NewHub(sessionID, idleTimeout)
+			testHub := NewHub(sessionID, idleTimeout)
 
-			Expect(hub.sessionID).To(Equal(sessionID))
-			Expect(hub.idleTimeout).To(Equal(idleTimeout))
-			Expect(hub.clients).To(BeEmpty())
-			Expect(hub.register).NotTo(BeNil())
-			Expect(hub.unregister).NotTo(BeNil())
-			Expect(hub.events).NotTo(BeNil())
-			Expect(hub.commands).NotTo(BeNil())
+			Expect(testHub.sessionID).To(Equal(sessionID))
+			Expect(testHub.idleTimeout).To(Equal(idleTimeout))
+			Expect(testHub.clients).To(BeEmpty())
+			Expect(testHub.register).NotTo(BeNil())
+			Expect(testHub.unregister).NotTo(BeNil())
+			Expect(testHub.events).NotTo(BeNil())
+			Expect(testHub.commands).NotTo(BeNil())
 		})
 	})
 
 	Describe("RegisterClient", func() {
 		It("should register a client with hub", func() {
-			hub := NewHub("test-session", time.Minute)
-			done := make(chan struct{})
-
-			go func() {
-				hub.Run()
-				done <- struct{}{}
-			}()
-
-			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-				upgrader := websocket.Upgrader{}
-				conn, _ := upgrader.Upgrade(w, r, nil)
-				defer func() { _ = conn.Close() }()
-			}))
-			defer server.Close()
-
 			dialer := websocket.Dialer{}
-			wsURL := "ws" + strings.TrimPrefix(server.URL, "http")
 			conn, _, err := dialer.Dial(wsURL, nil)
 			Expect(err).NotTo(HaveOccurred())
 			defer func() { _ = conn.Close() }()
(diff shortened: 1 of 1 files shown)


pkg/ws/hub.go
```diff
@@ -0,0 +1,84 @@
...
+func NewHub(sessionID string) *Hub {
+    h := &Hub{
+        sessionID: sessionID,
+        clients:   make(map[*Client]struct{}),
+        register:  make(chan *Client),
+        unregister: make(chan *Client),
+        events:    make(chan Message, 256),
+        commands:  make(chan Message, 32),
+    }
+    go h.Run()
```
Egzavyer: Why do we call this inside the constructor? Can't we just call it on the hub instance when we're ready to use it?

pkg/ws/client.go
```diff
@@ -0,0 +1,53 @@
...
+    return &Client{
+        conn: conn,
+        hub:  hub,
+        send: make(chan Message, 256),
+    }
+}
+
+func (c *Client) ReadPump() {
+    defer func() {
+        c.hub.Unregister(c)
```
Egzavyer: We unregister the client every time they read a message from the hub?

Egzavyer (dismissed):
Solid overall, just a couple things. Also needs tests and fix linter errors

internal/ws/hub.go
```diff
@@ -0,0 +1,84 @@
...
+    // Channels for register/unregister clients and broadcast msgs
+    register   chan *Client
+    unregister chan *Client
+    events     chan Message
+    commands   chan Message
+    
+    mu sync.RWMutex
+}
+
+func NewHub(sessionID string) *Hub {
```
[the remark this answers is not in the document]
xsachax: So here's what's going on. Each debug session instantiates a hub, and a hub can create up to x client connections. Hub and server were indirectly tied, since when a debug session, aka server, was shut down - it took out everything including the hub and the OS cleaned up behind it.

I added a clear gap between server and hub to make them independant, which solves the memory leak in the event that we want to host several debug sessions without shutting down the server.

internal/ws/hub.go
```diff
@@ -0,0 +1,84 @@
...
+    h := &Hub{
+        sessionID: sessionID,
+        clients:   make(map[*Client]struct{}),
+        register:  make(chan *Client),
+        unregister: make(chan *Client),
+        events:    make(chan Message, 256),
+        commands:  make(chan Message, 32),
+    }
+    go h.Run()
```
xsachax: Fixed: This ties into the comment above and assumed server and hub were more connected than they should have.
Hub is **not a singleton**.

pkg/ws/client.go
xsachax: defered func. ReadPump is called as a goroutine so it will only unregister at the end when the loop exists

internal/ws/protocol.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Helpers
+func (e EventType) Message() Message {
+    return Message{Type: string(e)}
+}
+
+func (c CommandType) Message() Message {
+    return Message{Type: string(c)}
+}
```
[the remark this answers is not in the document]
xsachax: Just realized this is dead code. Our actual events (GoroutineEvent, InspectResult) carry data and need full msg construction anyway

internal/ws/hub.go
```diff
@@ -0,0 +1,84 @@
...
+                delete(h.clients, client)
+                close(client.send)
+            }
+            h.mu.Unlock()
+            
+        case event := <-h.events:
+            h.mu.RLock()
+            for client := range h.clients {
+                select {
+                case client.send <- event:
```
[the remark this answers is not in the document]
Egzavyer: Oh right my bad, still getting used to the syntax

config/config.go
Egzavyer: Super clean 👍

cmd/bingo/main.go
```diff
@@ -157,7 +176,11 @@ func getSymbolTable(proc string) *gosym.Table {
 	if err != nil {
 		panic(err)
 	}
-	defer exe.Close()
+	defer func() {
+		if err := exe.Close(); err != nil {
```
Egzavyer: Pretty sure I wrote this but like you said we should have proper logging instead of just panic. I might merge a branch into yours fix these

internal/ws/hub.go
```diff
@@ -0,0 +1,133 @@
...
+		select {
+		case <-ticker.C:
+			// Check idle timeout
+			if h.idleTimeout > 0 && len(h.clients) == 0 {
+				if time.Since(h.lastActivity) > h.idleTimeout {
+					log.Printf("Session %s idle for %v, shutting down", h.sessionID, h.idleTimeout)
+					h.shutdown()
+					return
+				}
+			}
```
Egzavyer: Super clean

.devcontainer/setup.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -e  # Exit on error
+
+echo "🚀 Setting up development environment..."
+
+# Display Go version
+echo "📦 Go version:"
+go version
```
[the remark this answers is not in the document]
jwt2706: Or does `set -e` already cover all of this?

internal/ws/server.go
```diff
@@ -0,0 +1,96 @@
...
+
+type Server struct {
+	addr   string
+	hubs   map[string]*Hub
+	config config.WebSocketConfig
+	mu     sync.RWMutex
+}
+
+func NewServer(addr string) *Server {
+	return NewServerWithConfig(addr, config.Default().WebSocket)
```
jwt2706: What's the purpose for `NewServer()`? Why not just call `NewServerWithConfig()`? We're using default values for the config parameter anyway.

internal/ws/server.go
```diff
@@ -0,0 +1,96 @@
...
+	hub := s.GetOrCreateHub(sessionID)
+	client := NewClient(conn, hub, r.RemoteAddr)
+
+	go client.ReadPump()
+	go client.WritePump()
+
+	hub.Register(client)
+}
+
+func (s *Server) GetOrCreateHub(sessionID string) *Hub {
```
jwt2706: Functions should have one purpose. Get and Create should be separate functions for clarity and consistency.

jwt2706:
Are we merging this directly into main, or are we using a dev staging branch in our workflow still?

jwt2706 (changes requested):

internal/ws/hub.go
```diff
@@ -0,0 +1,133 @@
...
+				select {
+				case client.send <- event:
+				default: // if event buffer is full, unregister the client to protect hub
+					slowClients = append(slowClients, client)
+				}
+			}
+			h.mu.RUnlock()
+			for _, client := range slowClients {
+				log.Printf("Client %s is slow; unregistering from hub %s", client.id, h.sessionID)
+				h.Unregister(client)
```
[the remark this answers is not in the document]
xsachax: The typical lifecycle would be to attempt a reconnect from the client side. (using something like an exponential backoff). Theres nothing useful to do on the server side currently. We can't assume the client wants to reconnect if in most cases it's a dead client. Retrying messages or connections would just create a pile up in memory.

We'll keep this in mind when we develop our clients.

cmd/target/target.go
```diff
@@ -1,11 +1,11 @@
-package debugger
+package main
```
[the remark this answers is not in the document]
xsachax: Definitely. We can even factor these in with our integration tests later on. According to go's idiomatic pattern they would still fall under cmd/ I think.... Something to think about in the future.

internal/ws/server.go
xsachax: Generally I would agree for clarity, but in this case the operations are tied together using a single mutex. Having two functions would create a race condition - GetOrCreateHub is the lock holder.

Egzavyer (dismissed):
LGTM 👍

internal/ws/server.go
jwt2706: Makes sense. I would advocate for a different name maybe, but I can't think of anything better at the moment, so it's good for now.

jwt2706 (dismissed):
Awesome, LGTM 🚀

jwt2706 (dismissed):
Honestly all looks good. 🔥

In the future, we could add a shield.io badge in `README.md` to reflect test coverage, but that's not urgent.

Egzavyer (dismissed):
LGTM

Egzavyer (approved):
LGTM

PR #39, opened 2026-02-27
38 replace syscall package with golangorgxsys

Egzavyer:
Replace frozen `syscall` package with the actively maintained external `golang.org/x/sys/unix` package.


jwt2706 (approved):
LGTM

PR #41, opened 2026-02-28
40 Support Cross-Compilation

Egzavyer:
Adds backbone to enable cross-compilation

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -41,7 +41,9 @@
         "go.testTimeout": "30s"
       },
       "extensions": [
-        "golang.go"
+        "golang.go",
+        "aaron-bond.better-comments",
+        "usernamehw.errorlens"
       ]
     }
   }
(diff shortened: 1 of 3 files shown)


cmd/bingo/main.go
```diff
@@ -4,8 +4,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
-
-	"golang.org/x/sys/unix"
+	"syscall"
```
jwt2706: Are we switching back to using the syscall library? I thought this was deprecated.

internal/debuginfo/debug_info.go
```diff
@@ -14,63 +10,9 @@ type Target struct {
...
-
-func (d *DebugInfo) LineToPC(file string, line int) (pc uint64, fn *gosym.Func, err error) {
-	return d.SymTable.LineToPC(file, line)
-}
-
-func (d *DebugInfo) LookupFunc(fn string) *gosym.Func {
-	return d.SymTable.LookupFunc(fn)
+type DebugInfo interface {
+	GetTarget() Target
+	PCToLine(pc uint64) (file string, line int, fn *gosym.Func)
```
jwt2706: Like in `debugger.go`, there should be comments above each function in the interface that gives it a bit of an explanation for the sake of maintainability.

jwt2706 (changes requested):
Few thoughts and questions. Overall clean code :D

PR #46, opened 2026-03-07
43 macos debugger support (darwin, arm64)

xsachax:
Complete debugger implementation for darwin arm64. 

Still need to clear our the permission situation but we can run it with:
1. `sudo just build darwin arm64`
2. `just sign` _(which signs the binary with an ad-hoc signature to simulate the certificate `com.apple.security.get-task-allow`). This is something we definitely cannot forget once we roll out releases._
3. `sudo just run darwin arm64`


internal/debugger/debugger_darwin_arm64.go
```diff
@@ -0,0 +1,717 @@
...
+
+// exceptionLoop runs in its own OS thread and handles all Mach exceptions
+func (d *darwinARM64Debugger) exceptionLoop() {
+	runtime.LockOSThread() // Mach msg requires fixed thread
+	defer runtime.UnlockOSThread()
+
+	for {
+		select {
+		case <-d.EndDebugSession:
+			return
```
jwt2706: Couldn't this just be an `if` statement instead of `select`? We don't do anything with `default` and there's only one `case`.

internal/debugger/debugger_darwin_arm64.go
```diff
@@ -0,0 +1,717 @@
...
+
+	if d.task != 0 {
+		C.clear_debug_exception_ports(d.task)
+		C.mach_port_deallocate(C.get_mach_task_self(), d.task)
+		d.task = 0
+	}
+
+	select {
+	case d.EndDebugSession <- true:
+	default:
```
jwt2706: same here

jwt2706 (changes requested):
couple comments

jwt2706 (dismissed):

Egzavyer (approved):

jwt2706 (approved):

PR #65, opened 2026-07-01
Linux fix: single-step the exact trapped thread

xsachax:
## Root cause (multi-factor)

`StepOver` hung on **linux/amd64** because of **three independent, compounding defects** in the ptrace backend + engine. `Continue → BreakpointHit` masked all three; the step-over dance (restore original byte → single-step → reinstall trap → resume) is what exposed them. Each had to be fixed — none alone is sufficient (see the ablation table).

1. **RIP not rewound after the `INT3` trap.** After a software breakpoint fires, the thread stops with `RIP = bp.addr + 1`. The engine *reported* the rewound address but never wrote it back to the thread, so after restoring the original byte at `bp.addr` the resume ran from **one byte into** the original instruction.
2. **ptrace commands issued off the tracer thread.** On Linux a tracee is owned by the exact OS thread that forked it under `PTRACE_TRACEME`; ptrace calls from any other thread fail with `ESRCH`. The wait loop ran on a *different* OS thread, so the in-`Wait()` signal re-injections (SIGURG re-delivery, step re-issue) silently failed with `ESRCH` — the tracee was never resumed and the debugger **hung**.
3. **Only the initial thread was traced.** Once (1) and (2) were fixed, the Go scheduler migrated the goroutine that owns the breakpoint onto an **untraced** runtime thread. Its `INT3` delivered `SIGTRAP` to the tracee's *own* runtime instead of to us → `SIGTRAP: trace trap`, tracee dies.

## The fix (surgical — `internal/debugger/backend_linux_amd64.go` + `engine.go`)

- **Single dedicated tracer thread (`ptraceExec`).** Every ptrace *command* (fork/attach/continue/step/read/write/detach and the in-`Wait()` re-injections) is funneled through one OS-locked goroutine, mirroring Delve's `execPtraceFunc`. `wait4` stays on the wait-loop thread (legal from any thread in the tracer's group). Kills the silent `ESRCH`. → fixes (2).
- **Rewind the trapped thread's PC (`engine.rewindTrapPC`).** Writes `bp.addr` back to the trapped thread on amd64 before any resume; no-op on arm64 (`BRK` leaves PC at the trap). → fixes (1).
- **Trace every thread (`PTRACE_O_TRACECLONE`).** Plus Delve's clone/SIGSTOP handling: resume the parent on `PTRACE_EVENT_CLONE`; set options on and resume each new child from its initial `SIGSTOP`. → fixes (3).
- **Per-thread single-step disambiguation (`stepTID`).** With every thread now traced, only a `cause==0` `SIGTRAP` on the exact thread we stepped is `StopSingleStep`; a sibling's `SIGURG` (Go preemption) mid-step-over is continued with the signal re-injected instead of being mistaken for the step result. This makes the identity of the rewound / single-stepped thread provably the trapped one — the thread-selection centerpiece.

## Verification & ablation — which factor is individually necessary

The full fix passes the immutable gate. Each factor was then removed **one at a time** on a throwaway branch and re-run on the same `Debugger E2E` gate. **No single change is sufficient — all three are necessary:**

| Variant | RIP-rewind | TRACECLONE | Result on `Debugger E2E` (ubuntu-24.04) |
|---|:---:|:---:|---|
| **Full fix** (`882bbd1`) | ✓ | ✓ | ✅ `TestE2EBasic` PASS (2.57s) · `TestE2EChurn` 200×/`-race` PASS (22.21s) — [run 28492902992](https://github.com/bingosuite/bingo/actions/runs/28492902992) |
| **− RIP-rewind** | ✗ | ✓ | ❌ `TestE2EBasic` **livelocks** — PC left one byte past the BP, restore→single-step runs mid-instruction, engine then sees a cascade of unmatched "spurious" SIGTRAPs and walks the PC forward one byte at a time (`advancing PC past BRK … 0x…e6→e7→e8→e9`), never emitting the next `BreakpointHit` → 15s harness **TIMEOUT** — [run 28493205809](https://github.com/bingosuite/bingo/actions/runs/28493205809) |
| **− TRACECLONE** | ✓ | ✗ | ❌ `TestE2EBasic` **crashes at 4s** with `SIGTRAP: trace trap` — the goroutine migrates to an untraced thread whose INT3 hits the tracee's own runtime — [run 28493217529](https://github.com/bingosuite/bingo/actions/runs/28493217529) |
| Baseline `debugger-e2e-base` | ✗ | ✗ | ❌ `StepOver` hangs at the first step-over (silent `ESRCH`, nothing resumes) |

Answering the specific questions:

- **RIP-rewind is necessary but not sufficient.** Removing only it turns a green gate into a spurious-SIGTRAP livelock/timeout in `TestE2EBasic`. In isolation (without the tracer-thread fix) it can't help at all, because commands never reach the tracee (hang) — so it "moves the needle" only in combination with the tracer-thread routing.
- **`PTRACE_O_TRACECLONE` is required — and not only for churn.** Without it even `TestE2EBasic` dies at ~4s (`SIGTRAP: trace trap`) as soon as the main goroutine migrates onto an untraced thread. `TestE2EChurn` (`GOMAXPROCS(4)` + 8 `LockOSThread` goroutines) makes that migration relentless, but the dependency already shows up in basic.
- **Tracer-thread routing is required for anything to resume at all** — it fixes the original hang. It was isolated via the diagnostic instrumentation used to find the bug (the in-`Wait()` re-issues returned `ESRCH` from the wrong thread); it isn't a separate table row because reverting it is a whole-backend change, but the baseline row *is* exactly this defect.

Gate files (`e2e_integration_test.go`, `debugger-e2e.yml`) are untouched; the change is confined to `backend_linux_amd64.go` + `engine.go`. Local: `gofmt`/`goimports` clean, `GOOS=linux go vet -tags e2e` clean, darwin `-tags bingonative` build/vet/unit-tests + full-module tests pass.

## Delve references

- `pkg/proc/native/proc.go` — `execPtraceFunc` (single tracer-thread executor model)
- `pkg/proc/native/proc_linux.go` — `trapWaitInternal`, `addThread`, `ptraceOptions` (`PTRACE_O_TRACECLONE`, clone/SIGSTOP handling)
- `pkg/proc/native/threads_linux.go` — per-thread single-step / breakpoint restore, amd64 PC adjust after the trap

### DIFF
--- a/internal/debugger/backend_linux_amd64.go
+++ b/internal/debugger/backend_linux_amd64.go
@@ -7,24 +7,107 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
+	"sync"
 	"syscall"
 )
 
 func newBackend() Backend {
-	return &linuxBackend{}
+	return &linuxBackend{pt: ptraceThread()}
 }
 
 type linuxBackend struct {
+	pt          *ptraceExec
 	pid         int
 	stepping    bool // true after SingleStep; classifies the next SIGTRAP
+	stepTID     int  // the exact thread SingleStep was issued against
 	lastStopTID int
 }
 
-const linuxPtraceOptions = syscall.PTRACE_O_TRACEEXIT |
+// ptraceExec serializes every ptrace(2) command onto one dedicated OS thread.
+//
+// Linux ties a tracee to the specific "tracer" thread that established the
+// trace relationship — here the thread that forks the child under
+// PTRACE_TRACEME. Every later ptrace request (PTRACE_CONT, PTRACE_SINGLESTEP,
+// register/memory access, PTRACE_SETOPTIONS, PTRACE_DETACH) must be issued
+// from that same thread; from any other thread the kernel rejects it with
+// ESRCH. The engine's wait loop runs on a *different* OS thread than the one
+// that forked the tracee, so ptrace commands re-issued from inside Wait()
+// (the transparent SIGURG / SIGCONT / clone re-steps) used to fail with ESRCH
+// — an error the helpers silently swallowed — leaving the tracee parked and
+// wedging step-over forever. Funnelling every command through this single
+// locked thread guarantees they all originate from the tracer.
+//
+// wait4(2) is deliberately NOT routed here: any thread in the tracer's thread
+// group may reap the tracee, and blocking this thread inside wait4 would stop
+// it from servicing the very commands the wait loop needs. This mirrors
+// Delve's model (dedicated ptrace thread for commands; wait4 on regular
+// goroutines — see pkg/proc/native/proc.go execPtraceFunc and
+// proc_linux.go trapWait).
+type ptraceExec struct {
+	mu    sync.Mutex
+	reqCh chan func()
+	ackCh chan struct{}
+}
+
+var (
+	ptraceOnce sync.Once
+	ptraceInst *ptraceExec
+)
+
+// ptraceThread returns the process-wide ptrace command thread, creating it on
+// first use. Exactly one debug target is traced at a time in this codebase, so
+// a single shared tracer thread keeps the fork (in startTracedProcess) and all
+// of the backend's subsequent commands on the same thread.
+func ptraceThread() *ptraceExec {
+	ptraceOnce.Do(func() { ptraceInst = newPtraceExec() })
+	return ptraceInst
+}
+
+func newPtraceExec() *ptraceExec {
+	p := &ptraceExec{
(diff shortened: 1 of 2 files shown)


xsachax:
Superseded by #69 (base `57-refactor-debugger-foundation`). Your **stepTID** trapped-thread disambiguation was the one non-destructive win folded into #69 on top of the single-tracer-thread foundation — step-completion and SIGURG re-single-step are now gated on `tid == b.stepTID`. Closing this standalone branch; the fix lives on in #69.

PR #66, opened 2026-07-01
Linux fix: rewind RIP into register after INT3 breakpoint

xsachax:
## Root cause

**THE root cause is a missing PC-register rewind after the `INT3` software-
breakpoint trap on amd64.** A software breakpoint overwrites the target byte with
`INT3` (`0xCC`, 1 byte). When the CPU executes it, it traps *after* advancing
past the byte, leaving the trapped thread's `RIP` one byte too far:

```
                         RIP value
  original instr @ addr, before trap   →   addr
  after the INT3 trap fires             →   addr + 1     ← hardware leaves it here
  what a correct resume needs           →   addr         ← must rewind register to here
```

`engine.populateBreakpointStop` rewinds the *reported* `stop.PC` (via
`archRewindPC`) so the emitted event looks right, **but it never writes that
value back into the thread's `RIP` register.** So on resume,
`resumeFromBreakpoint` restores the original opcode at `addr` and then
single-steps starting from `addr+1` — the *middle* of the just-restored
instruction — corrupting the tracee and wedging `StepOver` /
continue-past-breakpoint indefinitely (the mid-instruction livelock).

**Why arm64/Darwin is immune:** `BRK #0` leaves `PC` *at* `addr` (the trap does
not advance PC), so the reported `stop.PC` already equals the register and no
rewind is needed — which is precisely why this bug is `linux/amd64`-only.

### Additional requirements the `-race` gate exposed

The rewind is necessary but not sufficient for `TestE2EChurn`, which runs the
tracee under `-race` (heavy goroutine/thread churn). Two further backend defects
surfaced only under that churn:

1. **Cross-thread ptrace `ESRCH`.** Linux binds a tracee to the exact OS thread
   that established the trace (the thread that forked/exec'd it). The engine's
   `waitLoop` runs on a *different* OS thread, so any ptrace **control** op it
   issued while absorbing internal stops (`PTRACE_CONT`/`SINGLESTEP`/
   `SETOPTIONS`) returned `ESRCH` and silently no-op'd, stalling resumes.

2. **Non-traced clone threads.** `PTRACE_O_TRACECLONE` was off, so Go-runtime
   clone threads were untraced. Under `-race` the breakpoint-owning goroutine
   migrates onto such a thread, executes the `INT3` there, and the `SIGTRAP`
   escapes ptrace → fatal `SIGTRAP: trace trap` → the tracee dies instead of
   reporting a hit.

## The fix

- **`engine.go` (centerpiece):** new `rewindTrapPC(tid, pc)` writes the rewound
  PC back into the trapped thread's `RIP` the moment a real breakpoint is
  matched in `handleStop`, before any restore + single-step or plain continue.
  Guarded by `regs.PC == pc`, so it is a **no-op on arm64/Darwin** (where `BRK`
  already leaves PC at `addr`). Mirrors Delve's
  `nativeThread.SetCurrentBreakpoint → setPC(bp.Addr)`, gated on
  `Arch.BreakInstrMovesPC()`.

- **`backend_linux_amd64.go`:** serialize every ptrace command on a single,
  OS-locked, process-wide tracer thread (`ptraceExec` / `ptraceThread()`, funnelled
  through `run(fn)`), so fork/exec and all later control ops share one tracer and
  never `ESRCH`. `wait4(2)` stays on the wait-loop thread (any thread in the tracer
  group may reap). Enable `PTRACE_O_TRACECLONE`; enroll clone children on their
  initial `SIGSTOP` and resume the parent on `PTRACE_EVENT_CLONE`. Track
  `stepTID` and classify a `cause==0` `SIGTRAP` as step-completion only when
  `tid == b.stepTID` (the goroutine can migrate across threads). `SIGURG`
  (Go preemption) is re-delivered transparently.

Surgical: only `engine.go` (+35) and `backend_linux_amd64.go` change. No changes
to the immutable acceptance test or gate workflow.

## Verification (CI, real ptrace)

`Debugger E2E` run **[28493876907](https://github.com/bingosuite/bingo/actions/runs/28493876907)**
(commit `bda917e`) — **`Linux amd64 E2E (ptrace)` job GREEN**:

```
✓ Linux amd64 E2E (ptrace) in 1m9s
  ✓ E2E basic (continue + step-over correctness)   # TestE2EBasic
  ✓ E2E churn (multi-thread robustness)            # TestE2EChurn
```

(The `Darwin arm64 E2E` job is `continue-on-error` and fails only on the
CI runner's missing `com.apple.security.cs.debugger` entitlement — irrelevant to
this linux fix.)

**Ablation — isolating the centerpiece:** disabling *only* the RIP rewind (with
the tracer-thread + `TRACECLONE` still in place) reintroduces the
mid-instruction livelock — run
[28493205809](https://github.com/bingosuite/bingo/actions/runs/28493205809)
fails at `Linux amd64 E2E (ptrace)` → `E2E basic`. Disabling *only*
`PTRACE_O_TRACECLONE` likewise fails
([28493217529](https://github.com/bingosuite/bingo/actions/runs/28493217529)).
So both factors are individually necessary, and the RIP rewind alone is required
to fix the core step-over correctness bug.

Local (darwin host): `gofmt`/`goimports` clean; `GOOS=linux GOARCH=amd64 go
build/vet ./...` (normal + `-tags e2e`) clean; `go test -tags bingonative -race
./internal/debugger/ ./internal/hub/` pass.

## Delve references

- `pkg/proc/native/threads_linux.go` — `(*nativeThread).SetCurrentBreakpoint` →
  `setPC(bp.Addr)` (the PC-rewind pattern).
- `pkg/proc/amd64_arch.go` / arch `BreakInstrMovesPC()` — gates the rewind to
  archs where the trap advances PC (amd64 yes, arm64 no).
- `pkg/proc/native/proc_linux.go` — `trapWait`/`addThread` and
  `PTRACE_EVENT_CLONE` handling; the `execPtraceFunc` single-tracer-thread model
  that serializes all ptrace calls on one OS thread.

### DIFF
--- a/internal/debugger/backend_linux_amd64.go
+++ b/internal/debugger/backend_linux_amd64.go
@@ -7,24 +7,107 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
+	"sync"
 	"syscall"
 )
 
 func newBackend() Backend {
-	return &linuxBackend{}
+	return &linuxBackend{pt: ptraceThread()}
 }
 
 type linuxBackend struct {
+	pt          *ptraceExec
 	pid         int
 	stepping    bool // true after SingleStep; classifies the next SIGTRAP
+	stepTID     int  // the exact thread SingleStep was issued against
 	lastStopTID int
 }
 
-const linuxPtraceOptions = syscall.PTRACE_O_TRACEEXIT |
+// ptraceExec serializes every ptrace(2) command onto one dedicated OS thread.
+//
+// Linux ties a tracee to the specific "tracer" thread that established the
+// trace relationship — here the thread that forks the child under
+// PTRACE_TRACEME. Every later ptrace request (PTRACE_CONT, PTRACE_SINGLESTEP,
+// register/memory access, PTRACE_SETOPTIONS, PTRACE_DETACH) must be issued
+// from that same thread; from any other thread the kernel rejects it with
+// ESRCH. The engine's wait loop runs on a *different* OS thread than the one
+// that forked the tracee, so ptrace commands re-issued from inside Wait()
+// (the transparent SIGURG / SIGCONT / clone re-steps) used to fail with ESRCH
+// — an error the helpers silently swallowed — leaving the tracee parked and
+// wedging step-over forever. Funnelling every command through this single
+// locked thread guarantees they all originate from the tracer.
+//
+// wait4(2) is deliberately NOT routed here: any thread in the tracer's thread
+// group may reap the tracee, and blocking this thread inside wait4 would stop
+// it from servicing the very commands the wait loop needs. This mirrors
+// Delve's model (dedicated ptrace thread for commands; wait4 on regular
+// goroutines — see pkg/proc/native/proc.go execPtraceFunc and
+// proc_linux.go trapWait).
+type ptraceExec struct {
+	mu    sync.Mutex
+	reqCh chan func()
+	ackCh chan struct{}
+}
+
+var (
+	ptraceOnce sync.Once
+	ptraceInst *ptraceExec
+)
+
+// ptraceThread returns the process-wide ptrace command thread, creating it on
+// first use. Exactly one debug target is traced at a time in this codebase, so
+// a single shared tracer thread keeps the fork (in startTracedProcess) and all
+// of the backend's subsequent commands on the same thread.
+func ptraceThread() *ptraceExec {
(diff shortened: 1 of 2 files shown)


xsachax:
Superseded by #69 (base `57-refactor-debugger-foundation`). The live-RIP rewind after INT3 is fully subsumed by #69's `rewindToBreakpoint` (same effect, plus a TID==0 guard and no-op identity rewind on arm64/Darwin). No behavioural gap left to graft. Closing as redundant.

PR #67, opened 2026-07-01
Linux fix: classify traps and forward signals on resume

xsachax:
## What's distinct about this fix

Every Linux `StepOver` solution has to solve the same two problems — the amd64
`INT3` PC-rewind and the cross-thread ptrace `ESRCH`. **This PR solves the
cross-thread problem with a `StopThreadEvent` round-trip on the tracer goroutine
plus explicit trap/signal classification keyed on `stepTID`**, rather than by
having the wait goroutine single-step / resume the trapped thread in place.

Concretely:

- **`Wait()` only *classifies*; it never resumes.** It labels each stop —
  single-step-completion `SIGTRAP`, `INT3` `SIGTRAP`, `PTRACE_EVENT` clone stop,
  clone-child `SIGSTOP`, or a real signal — using `stepTID` (the thread the step
  was issued against) to tell the step's own trap apart from a breakpoint or a
  churn thread's async signal. This is the disambiguation centerpiece.
- **Resumes are round-tripped back to the tracer goroutine.** Anything needing a
  ptrace resume (clone/thread-lifecycle stops, and signals on non-debug threads)
  is handed up as a new `StopThreadEvent`; the engine loop — the one
  `LockOSThread`'d goroutine that owns ptrace — resumes it via
  `ptraceResumer.ResumeThread` and re-arms the wait loop. Nothing outside that
  goroutine ever issues `PTRACE_CONT`/`SINGLESTEP`.

The alternative "just single-step the trapped thread from the waiter" avoids
`ESRCH` by construction; this design instead keeps a single serialized ptrace
owner and moves *events*, not *control*, across goroutines. That makes the
trap/signal classification explicit and centralizes every resume decision in the
engine, at the cost of one extra event round-trip per thread stop.

## Root cause (three compounding defects in `internal/debugger`)

1. **INT3 PC never rewound into the tracee.** On amd64 `INT3` leaves
   `RIP = bp.addr+1`; the engine computed the rewind only for the *reported*
   `StopEvent.PC` and never wrote it back. `resumeFromBreakpoint` then
   single-stepped from mid-instruction → `SIGSEGV`, which the `StopSignal` path
   swallowed (continued with signal 0) → infinite re-fault → **hang**. Hidden on
   arm64 (`archRewindPC` is identity there).
2. **Cross-thread ptrace `ESRCH`.** `wait4(-1,…,WALL)` reaps from any thread, but
   ptrace *control* is restricted to the tracer thread. Our per-resume `waitLoop`
   goroutine resuming there returned `ESRCH` and wedged the tracee.
3. **Untraced migration.** With only the main thread traced, when the
   goroutine-under-inspection migrated OS threads and hit the INT3 there, the
   `SIGTRAP` escaped the tracer and the Go runtime aborted
   (`trace/breakpoint trap`) → `ProcessExited`.

## The fix

**`backend_linux_amd64.go`**
- `PTRACE_O_TRACECLONE` (was `TRACEEXIT|TRACEEXEC`): every clone child is
  auto-traced, so a migrated breakpoint is always caught, never escaping (fixes 3).
- `Wait()` classification via `stepTID` (see above); issues no ptrace resume.
- `ResumeThread(tid, signal)` — new tracer-goroutine resume for clone/thread
  events and non-debug-thread signals; forwards the signal, never re-delivers a
  bare `SIGSTOP`.
- `ResumeSignal` — forwards synchronous faults (`SEGV/BUS/ILL/FPE/STKFLT`) on the
  re-issued single-step and defers async signals (`SIGURG`) via `delayedSignal`
  for the next continue (Delve's `singleStep()`/`resumeWithSig()` model). Signals
  are never swallowed.

**`engine.go`**
- `case StopThreadEvent:` resumes via `ptraceResumer.ResumeThread` on the engine
  goroutine and re-arms `waitLoop` — no session-state change, no event emitted.
- `rewindTraceePC(tid, addr)` writes the rewound PC back into the tracee at every
  software-breakpoint hit (fixes 1). Guarded by `archRewindPC`, so it is a no-op
  on arm64/darwin. Covers user breakpoints *and* the temporary
  `<stepover-next>`/`<stepout-return>` breakpoints.

**`backend.go`**
- `signalResumer` → `ptraceResumer` (`ResumeSignal` + `ResumeThread`); new
  `StopThreadEvent` reason; `StopEvent.Signal` documented for the forward path.

## Why darwin still builds (no stub needed)

`ResumeThread`/`ResumeSignal` were **not** added to the core `Backend` interface.
They live on a separate, optional **`ptraceResumer`** interface, and the engine
reaches them only through a runtime type assertion —
`if sr, ok := e.backend.(ptraceResumer); ok { … }`. The darwin backend
(`backend_darwin_arm64.go`) implements neither method and needs no stub: the
assertion is simply `false` there, and darwin's `Wait` never produces
`StopThreadEvent`/`Stepping` signals, so those branches are inert. The package
compiles cleanly on darwin (`go build -tags bingonative ./...`) with the darwin
backend completely untouched — this PR fixes Linux only.

## Verification

No local Linux ptrace is available (docker emulation is broken), so correctness
is proven by the immutable CI gate.

- **`Debugger E2E` → `Linux amd64 E2E (ptrace)`: GREEN** on the final commit.
  - `E2E BASIC PASS: 25 continue+stepover iterations`.
  - `E2E CHURN PASS: 200 continue+stepover iterations under thread churn`
    (8 churn goroutines; migration tolerated; no hang/exit/error).
  - Green run: https://github.com/bingosuite/bingo/actions/runs/28494208708
- Local: `gofmt` clean; `GOOS=linux GOARCH=amd64 go vet ./internal/debugger/` +
  `go build ./...` clean; darwin `go build -tags bingonative ./...` clean; unit
  tests for `internal/debugger`, `internal/hub`, `internal/server`,
  `pkg/protocol` pass.

The darwin E2E job is pre-existing flaky (documented parked-M race) and is
`continue-on-error`; A/B runs confirm these changes don't regress it. Out of
scope here. The E2E harness (`e2e_integration_test.go`, `debugger-e2e.yml`) is
untouched.

## Delve references

- `pkg/proc/native/proc_linux.go` — `trapWait` (clone-stop handling, `addThread`),
  `ptraceOptionsNormal = PTRACE_O_TRACECLONE`.
- `pkg/proc/native/threads_linux.go` — `(*nativeThread).resumeWithSig`,
  `singleStep` (signal forwarding on resume/step).
- `pkg/proc/native/proc_linux.go` `setCurrentBreakpoint` / `BreakInstrMovesPC` —
  `setPC(bp.Addr)` after an INT3 stop (the PC-rewind this fix ports to amd64).

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/internal/debugger/backend.go
+++ b/internal/debugger/backend.go
@@ -30,6 +30,22 @@ type pidSetter interface {
 	setPID(pid int)
 }
 
+// signalResumer is an optional Backend capability for platforms (Linux) where
+// ptrace resume commands are restricted to the tracer thread — the engine's
+// event-loop goroutine — and therefore cannot be issued from the wait
+// goroutine. When a real (non-trap) signal stops the tracee, Wait returns a
+// StopSignal event and the engine calls ResumeSignal from its own goroutine to
+// re-step or continue, forwarding or deferring the signal. Backends whose
+// ptrace is process-wide (Darwin/Mach) don't implement this; the engine falls
+// back to a plain continue for them.
+type signalResumer interface {
+	// ResumeSignal resumes tracee thread tid after a non-trap signal. When
+	// stepping is true the tracee was mid single-step and must be re-stepped
+	// (fault signals delivered on the step, async signals deferred to the next
+	// continue); otherwise the signal is forwarded on a plain continue.
+	ResumeSignal(tid, signal int, stepping bool) error
+}
+
 func setPID(b Backend, pid int) {
 	if ps, ok := b.(pidSetter); ok {
 		ps.setPID(pid)
@@ -52,6 +68,7 @@ type StopEvent struct {
 	Reason   StopReason
 	TID      int
 	PC       uint64
-	ExitCode int // StopExited only
-	Signal   int // StopSignal only
+	ExitCode int  // StopExited only
+	Signal   int  // StopSignal only
+	Stepping bool // StopSignal only: the signal interrupted an in-flight single-step
 }
--- a/internal/debugger/backend.go
+++ b/internal/debugger/backend.go
@@ -30,20 +30,28 @@ type pidSetter interface {
 	setPID(pid int)
 }
 
-// signalResumer is an optional Backend capability for platforms (Linux) where
+// ptraceResumer is an optional Backend capability for platforms (Linux) where
 // ptrace resume commands are restricted to the tracer thread — the engine's
 // event-loop goroutine — and therefore cannot be issued from the wait
-// goroutine. When a real (non-trap) signal stops the tracee, Wait returns a
-// StopSignal event and the engine calls ResumeSignal from its own goroutine to
-// re-step or continue, forwarding or deferring the signal. Backends whose
-// ptrace is process-wide (Darwin/Mach) don't implement this; the engine falls
-// back to a plain continue for them.
-type signalResumer interface {
-	// ResumeSignal resumes tracee thread tid after a non-trap signal. When
-	// stepping is true the tracee was mid single-step and must be re-stepped
-	// (fault signals delivered on the step, async signals deferred to the next
-	// continue); otherwise the signal is forwarded on a plain continue.
+// goroutine. When Wait observes a stop that needs a ptrace resume (a real
+// signal, or a clone/thread lifecycle event under PTRACE_O_TRACECLONE), it
+// hands the stop back to the engine, which calls back here from its own
+// goroutine. Backends whose ptrace is process-wide (Darwin/Mach) don't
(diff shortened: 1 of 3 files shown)


xsachax:
Superseded by #69 (base `57-refactor-debugger-foundation`). The trap-classification concern is handled on #69 via the single-tracer-thread model plus the stepTID disambiguation. The signal-forwarding-on-resume design here introduces an alternative architecture (new resumer/event interfaces + engine signal-dispatch rewrite) rather than a non-destructive add-on, so it was intentionally left out to protect the verified foundation. Closing.

PR #68, opened 2026-07-01
Linux fix: canonical clone/thread lifecycle tracing

xsachax:
## TL;DR

The step-over/continue hang is fundamentally a **thread-affinity** bug, and this
PR fixes it the way Delve does: **confine the entire tracee interaction — the
`fork`/exec, every ptrace request, `wait4`, and the state they touch — to a
single dedicated OS thread** (`ptraceThread`, Delve's `execPtraceFunc`). Two
other conditions the gate exposed — enabling `PTRACE_O_TRACECLONE` and rewinding
the amd64 RIP past `INT3` — are necessary co-requirements, but neither works
until ptrace runs on the tracer thread.

## Root cause

Linux ties every ptrace **control** op (`CONT`/`SINGLESTEP`/`SETOPTIONS`/`PEEK`/
`POKE`/`GETREGS`) to the OS thread that created the tracee — here, the thread
that forks the `PTRACE_TRACEME` child. The engine funnels its command-side ptrace
onto one locked goroutine (`engine.loop`, the forking thread), but `Backend.Wait`
runs on a **separate** `waitLoop` goroutine. So the clone / thread-lifecycle
resumes that `Wait` performs executed on the wrong thread and **silently
no-oped**: CI instrumentation showed `PTRACE_CONT` from the wait thread returning
`nil` while leaving both the leader (`6642`) and the freshly cloned child
(`6645`) in state `t` (traced-stopped). The whole group stayed stopped and
`wait4` blocked to the test timeout.

This is why a naive "just add `TRACECLONE`" attempt made the *first* `Continue`
hang, and why single-tid resume let the main goroutine migrate onto an untraced
LWP and wedge the step-over.

## Fix — one dedicated ptrace thread (the centerpiece)

`internal/debugger/backend_linux_amd64.go`, modeled on `go-delve/delve`
`pkg/proc/native` `execPtraceFunc`:

- A `ptraceThread` goroutine `runtime.LockOSThread()`s for the process lifetime
  and runs closures off a channel. **The fork/exec, every ptrace request,
  `wait4`, and the bookkeeping they read/write are all marshalled onto it** via
  `do(fn)`. `newBackend` publishes the handle so the package-level process hooks
  (`startTracedProcess`/`attachToProcess`/`killProcess`, invoked through the
  shared `process` type rather than as backend methods) fork and reap on that
  same thread — making it the tracer for **all** subsequent control ops.
- `Wait`'s entire `wait4` + clone-handling loop runs inside a single `do`
  closure, so its ptrace never leaves the tracer thread. This is what makes the
  resumes actually take effect.
- `StopProcess` deliberately stays a bare signal (not routed through the thread)
  so it can interrupt a blocked `wait4` rather than queue behind it.

### Co-requirement 1 — canonical clone tracing (`PTRACE_O_TRACECLONE`)

Set on launch and attach (Delve's `ptraceOptionsNormal`). All LWPs are tracked in
`threads map[int]*threadState`; `Wait` is a state machine — on
`PTRACE_EVENT_CLONE` resume the parent, register+resume each new child from its
own initial stop (order-independent vs. the parent event), drop exited threads,
keep `SIGURG` (Go async preemption) transparent; only a real breakpoint/
single-step trap, an unexpected signal, or process exit surfaces to the engine.
`ContinueProcess` resumes **every** stopped thread so goroutine migration can't
wedge the group.

### Co-requirement 2 — amd64 RIP rewind

`internal/debugger/engine.go` + `trap_{amd64,arm64}.go`. After `INT3` the x86 CPU
leaves `RIP` one byte **past** the trap; the engine computed the rewound PC for
breakpoint *lookup* but never wrote it back, so `resumeFromBreakpoint`
single-stepped from mid-instruction. `handleStop` now rewinds the trapping
thread's real PC to the breakpoint address, gated by `archBreakpointTrapMovesPC()`
— amd64 `true`, arm64 `false` (BRK leaves PC at the instruction, so **Darwin/arm64
is semantically untouched**).

## Verification

No local linux ptrace is available (docker emulation broken); verified via the
immutable `Debugger E2E` gate (`ubuntu-24.04`, `-race`).

**Green run: https://github.com/bingosuite/bingo/actions/runs/28494576533**

- `TestE2EBasic` — PASS (25 continue+stepover iterations)
- `TestE2EChurn` — PASS (200 continue+stepover iterations under thread churn)

Locally: `gofmt`/`goimports` clean; `GOOS=linux GOARCH=amd64 go build/vet` clean;
`GOOS=darwin GOARCH=arm64 go build -tags bingonative` clean;
`go test -tags bingonative -race ./...` passes.

## Delve references

- `pkg/proc/native/proc_linux.go`: `execPtraceFunc` / `handlePtraceFuncs` (the
  single ptrace-owning thread this PR mirrors); `ptraceOptionsNormal =
  syscall.PTRACE_O_TRACECLONE`; `(*nativeProcess).addThread`; `trapWaitInternal`
  (`PTRACE_EVENT_CLONE` → register + resume child/parent, thread-exit `delete`,
  unknown-thread `continue`, transparent signals).
- `pkg/proc/native/threads_linux.go`: per-thread `running` state that
  `ContinueProcess` mirrors.

Scope: surgical; only `internal/debugger` (linux backend + engine rewind co-fix +
trap arch helpers). The gate files (`e2e_integration_test.go`,
`.github/workflows/debugger-e2e.yml`) are untouched.

### DIFF
--- a/internal/debugger/backend_linux_amd64.go
+++ b/internal/debugger/backend_linux_amd64.go
@@ -7,25 +7,98 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
 	"syscall"
 )
 
+// ── Dedicated ptrace thread ──────────────────────────────────────────────────
+//
+// Linux ties every ptrace *control* request for a tracee (PTRACE_CONT,
+// PTRACE_SINGLESTEP, PTRACE_SETOPTIONS, PTRACE_PEEK/POKE, PTRACE_GETREGS, …) to
+// the specific OS thread that created the tracee: the thread that forked a
+// PTRACE_TRACEME child, or PTRACE_ATTACHed it. Issuing such a call from any
+// other thread does not resume/step the target — on our CI kernels it silently
+// returns success while leaving the thread ptrace-stopped, wedging the whole
+// group.
+//
+// The engine already funnels its command-side ptrace onto one locked goroutine
+// (engine.loop, which is also the thread that forks the tracee). But Backend.Wait
+// runs on a *separate* waitLoop goroutine. The clone / thread-lifecycle handling
+// Wait performs (resuming the parent and each freshly cloned LWP) therefore ran
+// on the wrong thread and never actually resumed anything.
+//
+// We fix this the way Delve does (execPtraceFunc, pkg/proc/native/proc_linux.go):
+// a single dedicated OS thread owns the tracee for its entire lifetime. The
+// fork/exec, every ptrace request, wait4, and the bookkeeping they touch are all
+// marshalled onto it. ptraceThread is that thread.
+type ptraceThread struct {
+	reqs chan func()
+}
+
+func newPtraceThread() *ptraceThread {
+	pt := &ptraceThread{reqs: make(chan func())}
+	ready := make(chan struct{})
+	go func() {
+		// Pin to one OS thread for the process lifetime and never unlock: every
+		// ptrace call and every wait4 for the tracee must originate here.
+		runtime.LockOSThread()
+		close(ready)
+		for fn := range pt.reqs {
+			fn()
+		}
+	}()
+	<-ready
+	return pt
+}
+
+// do runs fn on the ptrace thread and blocks until it completes. Calls serialize:
+// while Wait is parked in wait4 inside its own do() closure, later do() calls
+// queue behind it until the tracee stops. That matches how the engine drives the
+// process — resume, then wait for the next stop — so no legitimate flow contends.
+func (pt *ptraceThread) do(fn func()) {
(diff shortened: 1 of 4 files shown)


xsachax:
Superseded by #69 (base `57-refactor-debugger-foundation`). Clone tracing is already on #69 via `PTRACE_O_TRACECLONE` with per-thread SIGSTOP resume (no group-continue), and the `archBreakpointTrapMovesPC()` helper duplicates #69's existing `archRewindPC`. The resume-all `ContinueProcess` change is a resume-semantics shift, not non-destructive, so it was left out. Closing as redundant.

PR #69, opened 2026-07-01
Linux fix: unify ptrace onto a single tracer thread

xsachax:
## Root cause

On `linux/amd64` the engine could set a breakpoint and `Continue` to it, but the
first `StepOver` **hung**, and once that was unwedged the tracee later **crashed**.
Three independent low-level defects were in play:

1. **Cross-thread ptrace control ops (the step-over hang).** ptrace is
   thread-bound on Linux — every request for a tracee must be issued from the OS
   thread that became its tracer. The engine issued control ops from the engine
   loop's locked thread, but `Wait()`'s *internal* control ops (SIGURG
   re-delivery, clone/stop handling) ran on the wait goroutine's **separate**
   locked thread. Those cross-thread calls returned `ESRCH`, so the tracee was
   never resumed after the single-step and `StepOver` wedged.

2. **Missing PC rewind after a software breakpoint (correctness).** On amd64 the
   CPU advances `RIP` past the `INT3` before delivering the trap. The engine
   rewound PC only in its local `StopEvent` (for table lookup/display), never in
   the tracee's **live** register. Every resume path — plain continue after a
   sentinel step BP, and the restore→single-step→reinstall step-over dance —
   therefore restarted one byte into the original instruction, corrupting the
   tracee so it never reached the expected stop.

3. **Untraced cloned threads (crash on the second Continue).** Without
   `PTRACE_O_TRACECLONE`, Go-runtime worker threads created via `clone()` were
   not traced. When `time.Sleep` migrated goroutine 1 onto such a thread and it
   hit an `INT3`, the `SIGTRAP` was delivered to the Go runtime's signal handler
   (`fatal error: unexpected signal ... trace trap`) instead of the tracer,
   crashing the target.

## Fix

Centerpiece — **unify the ptrace lifecycle onto one dedicated locked OS thread**,
mirroring Delve's `execPtraceFunc`:

- New `tracerThread` in `backend_linux_amd64.go` owns a single `runtime.LockOSThread`'d
  goroutine. **Every** ptrace control op (`fork/exec`+`SETOPTIONS`, attach,
  detach, `CONT`, `SINGLESTEP`, `GET/SETREGS`, `PEEK/POKEDATA`, and `Wait()`'s
  internal resumes) is funnelled through `execPtrace(fn)`. `wait4` stays *off*
  that thread (legal from any thread of the tracer process). `fork`/`attach`/
  `kill` are routed by threading `Backend` through the platform free funcs
  (Darwin ignores it). Clean shutdown via `closeTracer()` from the engine loop's
  defer.
- **PC rewind:** `rewindToBreakpoint` writes the rewound PC back into the live
  register on every software-BP stop. No-op on arm64/Darwin (identity rewind).
- **Clone tracing:** enable `PTRACE_O_TRACECLONE` (inherited at the
  single-threaded execve stop, so it covers all later Go-runtime threads) and
  resume each new thread's initial `SIGSTOP` **individually**, without touching
  the rest of the group — avoiding the "parking the thread group" hazard that
  had kept clone tracing disabled.

### Consolidation (grafted from the sibling linux solutions)

This PR is the agreed consolidation of the five linux candidate fixes. Only one
change was non-destructive enough to fold in on top of the single-tracer-thread
foundation:

- **`stepTID` trapped-thread disambiguation (from the "exact trapped thread"
  solution).** `Wait4(-1, …, WALL)` can return a *sibling* thread's `cause==0`
  SIGTRAP (a real breakpoint) or a SIGURG while a single-step is in flight.
  Keying step-completion off `b.stepping` alone misclassifies that sibling stop
  as the step's completion (or re-single-steps the wrong thread). The step
  branch is now gated on `tid == b.stepTID` (recorded in `SingleStep`, cleared
  in `ContinueProcess`). The single-threaded path is unchanged (tid always ==
  stepTID during a step); only the concurrent-stop corner changes, mirroring
  Delve's per-stop trap-thread tracking.

The other candidates were **rejected as redundant or non-additive**: the
standalone PC-rewind fix is subsumed by `rewindToBreakpoint`; the trap-file
`archBreakpointTrapMovesPC()` helper duplicates the existing `archRewindPC`; and
the signal-forwarding/clone-lifecycle rewrites introduce an alternative
architecture (new resumer/event interfaces, resume-all semantics) rather than a
non-destructive add-on, so they were left out to protect this branch's verified
behaviour.

## Verification

- `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./...` — clean.
- `GOOS=linux GOARCH=amd64 go vet ./internal/debugger/` — clean; linux/amd64
  test binary compiles.
- Darwin host: `go build -tags bingonative ./...` and
  `go test -tags bingonative ./internal/debugger/` — pass (shared engine
  unaffected by the linux-only change and the `stepTID` graft).
- The single-tracer-thread foundation was previously validated green on real
  ubuntu ptrace (`TestE2EBasic` + `TestE2EChurn`, `-race`) while the temporary
  E2E harness existed; that harness has since been removed from the tree by
  project direction, so this PR carries **no** test-harness or CI changes.
  Actual linux ptrace **runtime** re-validation needs a linux CI/host (it cannot
  run on the darwin dev machine).

## Delve references

- `pkg/proc/native/proc_linux.go` — `execPtraceFunc`/`handlePtraceFuncs` funnel
  every ptrace call through one goroutine locked to a single OS thread; `wait4`
  runs off that thread.
- `ptraceOptionsNormal = PTRACE_O_TRACECLONE`; `trapWaitInternal` clone handling
  auto-attaches and resumes new threads per-thread.
- `pkg/proc/native/threads_linux.go` — per-stop trapped-thread tracking (the
  `stepTID` analogue).

## Integration note

Base is `57-refactor-debugger-foundation`, which already contains the darwin
backend fix. This branch was rebased cleanly on top of it: it threads a
`Backend` argument through the shared `process.launch/attach/kill` and
`startTracedProcess/attachToProcess/killProcess` signatures, so it also touches
`backend_darwin_arm64.go` — but only to add an unused `_ Backend` parameter
(behaviour unchanged; required for the package to compile). The PR is
**mergeable/clean**; the darwin signature touch is the only cross-platform line
and is a no-op.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,10 +118,16 @@ This is the most fragile code in the repo. Read this section before changing
 anything in [internal/debugger/](internal/debugger/).
 
 1. **All ptrace/Mach calls run on a single OS thread.** The engine event loop
-   (`engine.loop`) calls `runtime.LockOSThread()` and never unlocks. ptrace
-   on Linux is thread-specific — calling it from a different OS thread fails.
-   Public `Debugger` methods (`Continue`, `SetBreakpoint`, …) submit a closure
-   to `cmdCh`; the loop executes it. They don't make ptrace calls themselves.
+   (`engine.loop`) calls `runtime.LockOSThread()` and never unlocks. Public
+   `Debugger` methods (`Continue`, `SetBreakpoint`, …) submit a closure to
+   `cmdCh`; the loop executes it. They don't make ptrace calls themselves.
+   ptrace is thread-bound on Linux, so the linux backend goes further: it owns
+   a **dedicated tracer thread** (`tracerThread`) and funnels *every* ptrace
+   control op through `execPtrace`, because they must issue from the exact
+   thread that forked/attached the tracee. `wait4` is the one exception — legal
+   from any thread of the tracer process, so `Wait` runs it directly off the
+   tracer thread. On Darwin the ptrace/Mach calls run on the engine-loop thread
+   itself (Mach ports are task-wide). Mirrors Delve's `execPtraceFunc`.
 
 2. **`waitLoop` is a one-shot, locked goroutine.** Every time the process is
    resumed, a fresh `waitLoop` goroutine is started. It calls `Backend.Wait()`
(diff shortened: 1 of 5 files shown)


Egzavyer:
Works:  Set breakpoint, continue, step over

PR #71, opened 2026-07-01
Darwin fix: per-thread hardware single-step

xsachax:
## Summary

Fixes the intermittent hang in `TestE2EBasic` / `TestE2EChurn` on **darwin/arm64**. There turned out to be **two** distinct kernel-interaction bugs — the second only became visible once the first was fixed. The centerpiece is **per-thread ARM64 hardware single-step**; the second fix drains a **Mach task-level suspend-count leak**.

Scope is surgical: only `internal/debugger/{backend_darwin_arm64.go, mach_darwin_arm64.h, engine.go}`. The immutable gate (`e2e_integration_test.go`, `.github/workflows/debugger-e2e.yml`) is untouched.

---

### Primary bug — wrong-thread single-step (the reported hang)

Darwin `ptrace(PT_STEP)` is **per-process, not per-thread**. In XNU `bsd/kern/mach_process.c`, `PT_STEP` calls `thread_setsinglestep()` on `get_firstthread(task)` — it arms `MDSCR_EL1.SS` on the task's **first** thread, *never* the thread that actually trapped.

The old `SingleStep(tid)` ignored `tid` and issued `PT_STEP` on `b.pid`:

```go
// Darwin PT_STEP is per-process (by PID), not per-thread.
if err := ptrace(ptDarwinStep, uintptr(b.pid), 1, 0); err != nil { ... }
```

On a Go program the goroutine running `main` migrates across Ms/threads (it calls `time.Sleep`), so the breakpoint is frequently hit by a thread that is **not** `thread[0]`. The step-over sequence in `engine.resumeFromBreakpoint` then:

1. restores the original instruction bytes at `bp.addr` (trap removed),
2. `SingleStep(bpTID)` → `PT_STEP` arms SS on `thread[0]` (**wrong thread**),
3. the task resumes; the real BP thread runs **free** (no trap byte, no SS) past `bp.addr`,
4. `thread[0]` is idle/parked → never delivers a step trap,
5. `Wait()` blocks forever → **HANG**.

### Secondary bug — Mach task-level suspend-count leak (exposed after fixing #1)

Once stepping was thread-targeted, a rarer wedge remained (~2/12 churn runs). Empirical tracing (a temporary `task_info(TASK_BASIC_INFO)` + per-thread `thread_info` watchdog on the blocked `wait4`) caught it red-handed:

```
==== BINGO SS-HANG DUMP (wait4 blocked > 12s) ====
  TASK suspend_count=1
  tid ... suspend=0  STATIC PCs=[..same..]   <== step target, unfrozen but NOT running
  tid ... suspend=1  [frozen sibling]   (x5)
```

Every thread frozen, **all PCs static** (including a thread mid-`ADD` in a busy loop — definitively frozen, not merely descheduled), while `wait4` blocked forever. The task's Mach **suspend count was stuck at 1**.

Mechanism: the BSD ptrace `resume:` path performs at most **one** `task_resume` per `PT_CONTINUE`. But a multithreaded tracee under Go's **SIGURG async-preemption storm** (sysmon preempting the churn workload's busy goroutines) gets `task_suspend`'d **more than once** when several threads enter the stop path concurrently — while `wait4` reports only a single stop. One `PT_CONTINUE` then leaves the count at 1: the whole task stays frozen and the next `wait4` never returns.

---

## 2. The fix

**Per-thread hardware single-step** (`SingleStep`, mirrors LLDB debugserver `EnableHardwareSingleStep`):

- **Trapped thread is not `thread[0]`:** freeze every *sibling* with `thread_suspend`, arm **only `MDSCR_EL1.SS`** on the target via `thread_set_state(ARM_DEBUG_STATE64)` (the kernel sets `PSTATE.SS` itself in `arm_debug_set64` on return-to-user — programming CPSR.SS ourselves corrupted the step state machine), then `PT_CONTINUE`. Only the target is runnable, so only it advances one instruction and traps.
- **Trapped thread is `thread[0]`:** keep `PT_STEP` (kernel arms SS on exactly that thread), siblings frozen.
- Freezing siblings also silences the SIGURG storm for the duration of the step, which otherwise interrupts every resume before the target can retire its one instruction (a livelock).

**Task-suspend drain** (`resumeTraced`, wraps *every* resuming ptrace op):

```go
pre := b.taskSuspendCount()          // task is stopped → count is settled
ptrace(request, b.pid, addr, data)   // BSD ptrace does exactly one task_resume
for i := 1; i < pre; i++ { b.taskResume() }  // cancel the leaked surplus
```

Race-free: while the post-resume count is positive the task **cannot execute**, so no new suspend can interleave with the drain; and `pre==1` (a healthy single stop) drains nothing.

**Supporting hardening:**
- Bounded `KERN_ABORTED` retry on the `thread_get/set_state` and single-step cgo helpers (debugserver does the same — the kernel transiently returns `KERN_ABORTED` right after a step exception / while a `thread_suspend` settles).
- Mid-step-death handling: if the tracee is killed between trap-removal and the step, report a clean `ProcessExited` instead of an opaque Mach error (`engine.resumeFromBreakpoint` + `stepErr`/`processGone`).
- `killProcess` now resumes the tracee under ptrace delivering `SIGKILL`, so a ptrace-stopped tracee actually dies instead of blocking the harness until timeout.

---

## 3. Verification (local, darwin/arm64, authoritative)

Build/sign/run:
```sh
CGO_ENABLED=1 go test -tags 'e2e bingonative' -race -c -o /tmp/d2.test ./internal/debugger
codesign --sign - --entitlements entitlements.plist --force /tmp/d2.test
/tmp/d2.test -test.v -test.run TestE2EBasic -test.timeout 120s
/tmp/d2.test -test.v -test.run TestE2EChurn -test.timeout 200s
```

**Authoritative runs — clean binary (no diagnostics), `-race`, no `BINGO_E2E_DEBUG`:**

| Test | Result |
| --- | --- |
| `TestE2EBasic` | **16 / 16 GREEN** (≥10 required) |
| `TestE2EChurn` | **10 / 10 GREEN** (≥5 required) |

Zero hangs, zero timeouts, zero data races. Sibling debugger sessions were running the same E2E suite throughout (they `kill -9` targets by basename; the handful of `ProcessExited {"exitCode":-1}` observed in wider batches are that external SIGKILL, excluded as an environmental confound — never a debugger hang).

**Robustness campaign — instrumented binary with the `wait4` watchdog on:** ~59 churn + 15 basic additional runs, **zero** `SS-HANG` dumps after the drain was applied to all resume paths. On the broken baseline `TestE2EBasic` hung ~6/6 at `iters=25`.

Existing unit tests: `go test -tags bingonative -race ./internal/debugger` → `ok`.

> The diagnostic tracing/watchdog used to localize the secondary bug was removed before finalizing; only the fix remains.

---

## 4. References

- **LLDB debugserver** — `DNBArchMachARM64::EnableHardwareSingleStep` sets **only** `MDSCR_EL1.SS` (never CPSR) and retries thread-state accessors on `KERN_ABORTED`. This fix matches that exactly.
- **Delve** — `pkg/proc/native/threads_darwin.go`, `registers_arm64_darwin.go`: thread-targeted register/step handling via Mach thread state.
- **XNU** — `bsd/kern/mach_process.c` (`PT_STEP`/`PT_CONTINUE` operate on `get_firstthread`; one `task_resume` per resume), `osfmk/arm64/pcb.c` (`arm_debug_set64` sets `PSTATE.SS` from `MDSCR_EL1.SS`), `osfmk/arm64/status.c` (debug-state get/set).

### DIFF
--- a/internal/debugger/backend_darwin_arm64.go
+++ b/internal/debugger/backend_darwin_arm64.go
@@ -404,7 +404,26 @@ func (b *darwinBackend) Threads() ([]int, error) {
 func (b *darwinBackend) Wait() (StopEvent, error) {
 	for {
 		var ws syscall.WaitStatus
-		tid, err := syscall.Wait4(b.pid, &ws, 0, nil)
+		// Reactive task-suspend drain — the race backstop for resumeTraced's
+		// resume-time drain. Poll for a pending stop without blocking. If none
+		// is reapable yet the task is still Mach-suspended, a surplus
+		// task_suspend leaked past the resume-time drain (resumeTraced reads the
+		// count once, before the resume, so a suspend from the same stop cluster
+		// that lands just after that read is missed). A Mach-suspended task
+		// cannot run to create a new stop, and WNOHANG proves none is queued, so
+		// any positive count here is pure leaked surplus: drain one and re-poll
+		// until the task is runnable again or a real stop surfaces. Bounded so a
+		// misbehaving task_resume can never spin forever.
+		tid, err := syscall.Wait4(b.pid, &ws, syscall.WNOHANG, nil)
+		for drains := 0; err == nil && tid == 0 && drains < 64 && b.taskSuspendCount() > 0; drains++ {
+			if rerr := b.taskResume(); rerr != nil {
+				break
+			}
+			tid, err = syscall.Wait4(b.pid, &ws, syscall.WNOHANG, nil)
+		}
+		if err == nil && tid == 0 {
+			tid, err = syscall.Wait4(b.pid, &ws, 0, nil)
+		}
 		if err != nil {
 			if isNoChildProcess(err) {
 				return StopEvent{Reason: StopExited, TID: b.pid}, nil
@@ -613,10 +632,14 @@ func (b *darwinBackend) taskResume() error {
 //
 // We read the suspend count while the task is still stopped, issue the request,
 // then perform exactly (pre-1) extra task_resume calls to cancel the leaked
-// suspends. This is race-free: whenever the post-request count is still positive
-// the task cannot execute, so no new suspend can interleave with the drain, and
-// we never resume past the single legitimate suspend of a healthy stop (pre==1
-// drains nothing).
+// suspends. This handles the common case cheaply at resume time, but it is NOT
+// sufficient on its own: `pre` is sampled once, before the request, so a surplus
+// suspend from the same stop cluster that lands just after the read is missed and
+// the count sticks at 1. The reactive drain-before-block in Wait() is the
+// authoritative backstop that closes that race; this drain merely keeps the
+// common case off the reactive path. (It cannot over-drain: XNU serializes ptrace
+// stops per process, so a positive count never covers a second, independently
+// wait4-reportable stop — only leaked surplus.)
 func (b *darwinBackend) resumeTraced(request, addr, data uintptr) error {
 	pre := b.taskSuspendCount()
 	if err := ptrace(request, uintptr(b.pid), addr, data); err != nil {
--- a/internal/debugger/backend_darwin_arm64.go
+++ b/internal/debugger/backend_darwin_arm64.go
@@ -35,11 +35,33 @@ type darwinBackend struct {
 	stepMu   sync.Mutex
 	stepping bool
 	stepTID  int // Mach thread port saved by SingleStep, reused on the step trap
+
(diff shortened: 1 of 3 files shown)


xsachax:
Superseded by #70 (surviving darwin solution). Same SIGURG re-injection horn as #70 but a larger mechanism swap; #70 is the consolidated darwin fix. Closing in favor of #70.

PR #72, opened 2026-07-01
Darwin fix: race-free wait/kill and signal delivery

xsachax:
## Summary

Makes the **darwin/arm64** debugger drive the ad-hoc-signed E2E targets reliably: no StepOver/Continue hang, no wedged `Kill`, no AMFI `SIGKILL`. All changes are surgical and confined to `internal/debugger` (plus an AGENTS.md update). The immutable gate (`e2e_integration_test.go`, `debugger-e2e.yml`) is untouched.

The bug was a cluster of three defects in the **wait / stop / continue / signal** state machine, plus a wedge-prone `Kill`.

### 1. Async-signal re-injection broke Go preemption — the systematic 6/6 StepOver hang
Go async preemption is **thread-directed**: the runtime `pthread_kill`s `SIGURG` (16) at the *specific* M whose goroutine must reach a GC safe point. The old resume path re-injected that signal through the ptrace resume-signal argument (`ptrace(PT_CONTINUE, pid, 1, SIGURG)`). XNU then delivers it **process-wide** (`psignal`) to an arbitrary thread — the intended M never runs its handler, its `signalPending` stays set, `runtime.preemptM` never re-sends, and a stop-the-world hangs forever.

### 2. Software breakpoints tripped AMFI — the `SIGKILL` deaths
The target is an ad-hoc-signed Go binary. Patching its `__TEXT` with a `BRK` invalidates the code signature; on the next page-in AMFI SIGKILLs the tracee (*"has no CMS blob? … Unrecoverable CT signature issue, bailing out"*). `PT_ATTACH` clears `CS_KILL` but does not stop the pager-level kill, and W^X blocks RWX. Memory patching is fundamentally unsafe here.

### 3. The mid-run re-arm timer deadlocked the Go runtime — the residual rare churn hang
Hardware breakpoints live in each thread's debug registers, so a Go M created *between* two stops (the churn target spawns them constantly via `runtime.LockOSThread`) runs unarmed and misses the breakpoint. The fix for that — a timer that re-arms while the process runs — was implemented by **Mach-suspending the whole task every 10 ms** to force threads off-core so `thread_set_state` reliably lands. Freezing sysmon / the netpoller / a scheduler-or-timer-lock holder at that cadence occasionally **loses `main`'s `time.Sleep` wakeup**, parking every M — all correctly armed — with none ever reaching the breakpoint. Ring-buffer capture proved this is a runtime lost-wakeup deadlock, not a missed breakpoint.

### 4. `Kill` could wedge
A ptrace-stopped tracee ignores `SIGKILL` until resumed, and a Mach-suspended thread (from an in-flight single-step, or the suspend-ticker above) blocks termination — so `Kill` did not return within the harness's 5 s window.

## Fix

- **Resume with signal 0.** `ContinueProcess` and `resumeAfterSignal` never re-inject async signals; the kernel delivers the still-pending, thread-directed signal to its intended M. Matches Delve's `ptraceCont`, which always passes 0.
- **Hardware breakpoints instead of code patching.** Breakpoints are armed in the AArch64 debug registers (`DBGBVR`/`DBGBCR` via `ARM_DEBUG_STATE64`, `bingo_set_thread_hw_breakpoints`). No memory is patched → signature stays valid, no i-cache issue. `WriteMemory` transparently maps the engine's trap/original pokes onto `armHWBreakpoint`/`disarmHWBreakpoint`, so the **cross-platform engine is unchanged**. `applyDebugState` re-arms every thread on each resume; `Wait` resolves a SIGTRAP via `threadAtHWBreakpoint` (the thread whose PC == an armed address).
- **Non-suspending mid-run re-arm.** `rearmWhileRunning` now re-applies the set with a plain per-thread `thread_set_state` (no `task_suspend`). Arming-on-resume is the reliable path (the tracee is ptrace-stopped, so every existing thread is off-core); the timer only needs to catch newly-created, still-yielding M's, which context-switch within microseconds and pick up the debug registers from an ordinary write. This removes the deadlock **and** the teardown wedge the suspend caused, so `armMu`/`killing`/`beginTeardown` are deleted.
- **Deterministic bounded `Kill`.** `killProcess` drains every thread's suspend count, delivers `SIGKILL` by every avenue (`PT_CONTINUE(SIGKILL)` → out-of-band `SIGKILL` → `PT_KILL` → `PT_DETACH(SIGKILL)`), then reaps with a 3 s bound so a stuck tracee can never hang the engine loop past the 5 s window.
- **PT_STEP is per-process on darwin** (arms the task's first thread), so an arbitrary thread cannot be single-stepped. `needsTempBPStepOver()` makes the engine step over a breakpoint with a transient next-line breakpoint + continue instead of the restore→single-step→reinstall dance.

## Verification

Local darwin/arm64 (authoritative), `-race`, codesigned with `entitlements.plist`, **without** `BINGO_E2E_DEBUG`:

```
CGO_ENABLED=1 go test -tags 'e2e bingonative' -race -c -o /tmp/d5.test ./internal/debugger
codesign --sign - --entitlements entitlements.plist --force /tmp/d5.test
```

| Suite | Result |
| --- | --- |
| `TestE2EBasic` × 12 | **12 GREEN** — 0 hang, 0 SIGKILL, 0 Kill-wedge |
| `TestE2EChurn` × 14 (200 iters each) | **11 GREEN** (including a clean **6/6 consecutive** streak, ~95–106 s each), 0 hang, **0 Kill-wedge**; 3 early (13–25 s) external `SIGKILL`s |
| `go test -tags bingonative -race ./internal/debugger` | ok |

No `"Kill did not return within 5s"` / `"backend may be wedged"` warning in any of the 26 runs.

**About the 3 SIGKILLs:** all struck early (13–25 s, i.e. mid-loop, not the 600 s timeout), reported as `ProcessExited {"exitCode":-1}` (death by signal, not a debugger hang). They occurred only while **concurrent debugger-E2E sessions were sharing the machine** (load ≈ 4–5; sibling `d1.test`/`d4.test`/`dprobe.test` + their churn targets). Memory was **86 % free**, ruling out OOM/jetsam; with HW breakpoints the target's code is never patched, so this is environmental cross-session interference, not a fault in this change. The clean 6/6 consecutive churn streak landed once that load dropped.

## Delve references

- `pkg/proc/native/proc_darwin.go` — legacy macnative backend; confirms the darwin thread-notification gap (*"Cannot be notified of new threads on OS X"*), i.e. debug registers must be re-applied on every stop, which is exactly what `applyDebugState` does.
- `pkg/proc/native/ptrace_darwin.go` — `ptraceCont` always resumes with signal `0`; the basis for the "resume with signal 0" fix.
- Modern Delve delegates darwin to `debugserver`/LLDB, which likewise arms hardware debug registers per-thread rather than patching ad-hoc-signed code — corroborating the HW-breakpoint approach on Apple Silicon.

---

## Update — two-layer cross-check, Delve HEAD audit, and low-load re-verification

A parallel investigation framed darwin as a two-layer failure. Mapping this fix onto that framing, and re-verifying at low load:

### Which layers this fix closes — both

- **Layer 1 (deterministic step-over hang — per-process `PT_STEP` steps the wrong thread): closed by avoiding `PT_STEP` entirely.** `needsTempBPStepOver()` makes the engine step over with a transient next-line **hardware** breakpoint + plain continue. There is no arbitrary-thread single-step in the step-over path, so the wrong-thread `PT_STEP` cannot occur — and no thread is Mach-suspended mid-step (itself a lost-wakeup risk).
- **Layer 2 (residual churn wedge — `SIGURG` misdirection): closed by resuming with signal 0, preemption left ON.** Go sends `SIGURG` thread-directed (`pthread_kill` at the target M), so it is already pending on the correct thread; resuming with `0` lets the kernel deliver it there naturally, and the handler clears that M's `signalPending`. Never re-posting it process-directed avoids both (a) a stuck `signalPending` and (b) a spurious process-wide `SIGURG` interrupting an M parked in `__psynch_cvwait`. **`GODEBUG=asyncpreemptoff` is not set** — async preemption stays fully enabled.

### Delve HEAD audit (`d716866`) — why "don't re-inject" is the thread-correct fix

- `pkg/proc/native/ptrace_darwin.go`: `ptraceCont(tid, sig)` **ignores `sig` and always passes `0`**.
- Stop source is **Mach exceptions** (`task_set_exception_ports(EXC_MASK_BREAKPOINT|EXC_MASK_SOFTWARE)`); `Wait4` is used only for exit reaping.
- **`PT_THUPDATE` is never called** anywhere in Delve (only the vendored `x/sys` constant exists).

`PT_THUPDATE` re-delivery was considered and rejected: the signal is *already* pending thread-correctly, so re-injecting it (by any avenue) produces a spurious second `SIGURG` — reintroducing the Layer-2 `__psynch_cvwait` interruption. The reference debugger's policy is exactly the one used here: resume with `0`, never re-inject.

### Low-load re-verification (rebuilt from HEAD, `-race`, codesigned, no `BINGO_E2E_DEBUG`, sequential `-test.count=1`)

| Suite | Genuine result | Excluded (environmental) |
| --- | --- | --- |
| `TestE2EBasic` × ~15 | **13 GREEN**, 0 hang, 0 DATA RACE, 0 Kill-wedge | 5 runs where the **target** died at ~0.7 s with `exitCode:-1` (external SIGKILL) |
| `TestE2EChurn` × ~28 (200 iters) | **every genuine completion GREEN**; 0 hang, 0 DATA RACE, 0 Kill-wedge | target-kills at *random* points (StepOver #8/#18, Continue #98 ⇒ external, not a fixed deadlock) + rc=137 `Killed: 9` of the `.test` binary itself under load 4+ |

Random kill points and whole-`.test`-binary `SIGKILL`s are the signature of cross-session name-based kills, not a debugger deadlock. **Zero genuine hangs, DATA RACEs, or `"Kill did not return"` warnings across every run, this batch or the committed streak** (which included a clean 6/6 consecutive churn). A longer fresh consecutive churn streak will be appended once the shared machine is quieter.

---

## XNU source-level confirmation of the Layer-2 negative result

To make the "don't re-inject; thread-directed re-delivery is impossible under wait4" claim rigorous (not just empirical), I verified it against XNU source (`bsd/kern/mach_process.c`, `bsd/kern/kern_sig.c`):

- **`PT_STEP` is per-process (Layer 1 root cause).** `ptrace()` PT_STEP/PT_CONTINUE take `th_act = get_firstthread(task)` and call `thread_setsinglestep(th_act, …)` — the *first* task thread, not the breakpoint thread. Confirms why arbitrary-thread single-step must be avoided; `needsTempBPStepOver()` does exactly that.
- **`PT_THUPDATE` cannot resume a wait4 stop.** It sets `ut->uu_siglist |= sigmask(data)` on the port-named thread and marks `p_xstat`/`SRUN`, but has **no `wakeup(&t->sigwait)` and no `task_resume`** — those live only under the PT_CONTINUE/PT_STEP `resume:` label. So it can flag the signal on the right thread but cannot drive delivery. Combined with `PT_CONTINUE(pid,0)`, the still-`P_LTRACED` process just re-enters the trace-stop for the re-added bit (a re-stop loop, not delivery).
- **`sig=0` cleanly drops the intercepted signal.** In `issignal_locked`, the wait4 trace-stop path clears the pending bit (`ut->uu_siglist &= ~mask`) *before* stopping; on resume `signum = p->p_xstat`, and with `p_xstat==0` the code `continue`s with the bit already gone. No re-stop, no spurious delivery.
- **`PT_CONTINUE(pid, SIGURG)` is the corrupting path.** It does `psignal(t, data)` — **process-directed** — *plus* a `p_xstat` re-add to the sigwait thread. Net: the intended M **plus** a spurious process-directed copy on another thread. That spurious copy is precisely the `__psynch_cvwait` interruption / lost-wakeup observed under churn. There is no wait4 primitive that delivers a thread-directed signal to exactly one thread.

**Conclusion.** The sibling's narrow claim is correct at the source level: under wait4 you cannot cleanly re-deliver a thread-directed signal (`PT_THUPDATE` has no resume; `PT_CONTINUE(sig)` double-delivers). But the conclusion that this forces `GODEBUG=asyncpreemptoff` is refuted: resuming with `0` **eliminates the Layer-2 failure mode by construction** (no re-injection ⇒ no spurious process-directed `SIGURG`) while leaving async preemption **enabled**, with no observer-effect GODEBUG imposed on the target. The only residual is a dropped async-preempt `SIGURG` leaving that M's `signalPending` stuck at 1 (`runtime.preemptM` only re-sends on `signalPending.CAS(0,1)`; reset happens solely in the handler's `doSigPreempt`) — benign for any target that reaches cooperative safe points, which the E2E churn target does constantly (its only call-free loop is the µs-scale 2000-iter `x += int64(i)`, far under the ~10 ms async-preempt threshold and bracketed by `atomic.AddInt64`/`UnlockOSThread`/`Sleep`). Fully-robust thread-directed delivery would require switching the stop source to Mach exception ports (`P_LSIGEXC`), which AGENTS.md places out of scope; logged as future work.

### Additional low-load churn streak (renamed binary to dodge cross-session `.test` name-kills)

Rebuilt-from-HEAD `-race` binary, codesigned, no `BINGO_E2E_DEBUG`, sequential `-test.count=1`, 200 iters, load ≈ 3:

| Run | Result | Elapsed |
| --- | --- | --- |
| 1, 2, 4 | excluded — child `churn_target` externally `SIGKILL`ed (`exitCode:-1`) | — |
| 3 | **PASS** | 94.2 s |
| 5 | **PASS** | 107.5 s |
| 6 | **PASS** | 104.2 s |

**3/3 genuine completions GREEN, 0 hang, 0 DATA RACE, 0 Kill-wedge.** Aggregated with the committed run and the earlier low-load batch, every genuine `TestE2EChurn` completion has passed with zero wedges — the ~15%-per-session re-injection residual is eliminated in practice, preemption left on.

---

## Empirical async-preemption proof — the load-bearing `sig=0` claim, tested not reasoned

A reviewer correctly flagged that the acceptance gate **cannot** validate the crux of this fix. `resumeAfterSignal` resumes an intercepted `SIGURG` with `sig=0`; the safety of that hinges on what happens to the dropped signal. But `TestE2EChurn`'s target **cooperatively yields every iteration**, so it reaches GC safe points on its own and passes *whether or not* `SIGURG` is delivered. The gate is blind to this exact question.

I closed the gap with a dedicated, env-guarded diagnostic probe: **`internal/debugger/asyncpreempt_e2e_test.go`** (build-tagged `e2e && darwin && arm64 && bingonative`, gated behind `BINGO_ASYNCPREEMPT_TEST` so it **never runs in the gate**). It pairs a goroutine in a genuinely **non-cooperative** tight loop — `for i { x = x*3 ^ i; sink = x }`, no calls / allocations / channel ops, hence no compiler-inserted safe point, *only* async-preemptible via `SIGURG` — with a `main` that runs a stop-the-world `runtime.GC()` every iteration. If STW can't preempt the looper, the whole process (including `main`) freezes and the debugger stops seeing breakpoint hits.

### Result (instrumented, not reasoned)

| Scenario | Under `sig=0` debugger | Verdict |
| --- | --- | --- |
| **Native un-traced** tight loop + STW GC | GC-PROGRESS climbs freely (iter≥9000 in ~10 s) | baseline: async preemption works |
| **Non-cooperative** tight loop (`TestDarwinAsyncPreemptGC`) | **HANG at iter #1** — `runtime.GC()` never returns (15 s waitFor timeout, reproduced 3×) | **`sig=0` DROPS the SIGURG** |
| **Cooperative control** (`TestDarwinAsyncPreemptGCControl`, non-leaf `//go:noinline` safe point) | **PASS** every iteration | cooperative code preempts fine |
| **Main-only GC baseline** (`TestDarwinGCMainOnly`, no busy goroutine) | **PASS** | STW GC itself is healthy under the debugger |

**Conclusion: the earlier "signal stays pending and is delivered naturally" comment was wrong — `sig=0` drops the intercepted `SIGURG`.** I have corrected that comment, the two related comments, and the AGENTS.md invariant to state this honestly (commit `62169b0`).

### It is a fundamental wait4 tradeoff, not a fixable bug

I also ran the opposite policy (re-inject the signal via `PT_CONTINUE(sig)`) on an otherwise-identical build:

| Policy | Non-cooperative tight loop | Cooperative / gate-like code |
| --- | --- | --- |
| **`sig=0`** (this PR) | HANG (deterministic — signal dropped) | **PASS** — 0 wedge across dozens of runs |
| **re-inject `SIGURG`** | PASS (`p_xstat` re-add reaches the target M) | **~17% WEDGE** (1/6 runs) — spurious process-directed `psignal()` copy interrupts an M in `__psynch_cvwait` = the original churn hang |

Neither wait4 policy satisfies both columns. Re-injection's ~17% cooperative wedge **matches the ~15%-per-session residual measured independently by the coordinator**, and it **fails the acceptance gate**. `sig=0` is therefore the correct choice: it is 100% reliable for every cooperative target — which is every realistic Go program and the entire E2E suite — at the cost of a corner case (a goroutine that is *only* async-preemptible won't be STW-preempted while traced). Thread-directed re-delivery is impossible under wait4 (`PT_THUPDATE` has no `wakeup(&t->sigwait)`/`task_resume` to drive the stop's resume; `PT_CONTINUE(sig)` is process-directed). The only complete fix is the **Mach-exception stop model**, where `SIGURG` is never intercepted (what Delve does) — explicitly out of scope per AGENTS.md.

This makes `sig=0` **strictly better than the sibling's `GODEBUG=asyncpreemptoff=1`**: that disables async preemption globally (breaking the tight loop *even un-traced*), whereas `sig=0` leaves preemption fully **on**, so the corner case is scoped to *non-cooperative goroutines while a breakpoint is being serviced*, and vanishes the instant the target has any safe point.

### Cross-reference: defect-3 ↔ D2 (#71) — same task-suspend hazard, two sources

My defect-3 (the debug-register re-arm ticker must **not** `task_suspend` the whole task every 10 ms — it froze sysmon / the netpoller / the timer-lock holder and lost `main`'s `time.Sleep` wakeup, deadlocking with every M parked-but-armed) is the **same `task_suspend` lost-wakeup hazard that D2 (#71) independently hit on the resume path**. Two darwin investigations triangulated the identical XNU pitfall — whole-task suspension interacting with `__psynch_cvwait`-parked runtime threads — from different entry points. The remedy is shared in spirit: never suspend the whole task to force a state transition; arm on resume (per-thread) instead.

### Fresh low-load re-verification after the doc-correction commit (`62169b0`)

The commit is comment/doc-only in the backend (the `ptrace(request, b.pid, 1, 0)` resume is byte-identical) plus the new env-guarded probe, so gate behavior is unchanged. Confirmed on a quiet machine (load ≈ 2.0), rebuilt `-race`, codesigned, no `BINGO_E2E_DEBUG`, sequential:

- **`TestE2EBasic` (iters=25): 10 / 10 GREEN** — 0 wedge, 0 DATA RACE, 0 sibling-kill, 0 Kill-wedge.
- **`TestE2EChurn` (iters=25): 6 / 6 GREEN** — 0 wedge, 0 DATA RACE, 0 sibling-kill, 0 Kill-wedge.
- Linux/amd64 cross-compile clean (`GOOS=linux GOARCH=amd64 go build ./internal/debugger`).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -236,17 +236,36 @@ engine advances PC by `len(archTrapInstruction())` and resumes. See the
   registers from a plain write), dropping the suspend removes the deadlock
   without reintroducing missed breakpoints. The timer is skipped while
   single-stepping.
-- **Do NOT re-inject async signals — resume with signal 0.** This is the crux of
-  correct Go-preemption support. Go async preemption is *thread-directed*: the
-  runtime `pthread_kill`s `SIGURG` (16) at the exact M whose goroutine must reach
-  a GC safe point. If you re-post it via `ptrace(PT_CONTINUE, pid, 1, SIGURG)`,
-  XNU delivers it PROCESS-wide (`psignal`) to an arbitrary thread; the intended M
-  never runs its handler, its `signalPending` stays set, `runtime.preemptM` never
-  re-sends, and a stop-the-world hangs forever (the original 6/6 StepOver hang).
-  `resumeAfterSignal` therefore resumes with signal 0 (`ptrace(request, pid, 1,
-  0)`) for `SIGURG`/`SIGWINCH`, letting XNU deliver the still-pending
-  thread-directed signal to its intended M. This matches Delve, whose
-  `ptraceCont` always passes 0.
+- **Do NOT re-inject async signals on darwin — resume with signal 0.** Go async
+  preemption is *thread-directed*: the runtime `pthread_kill`s `SIGURG` (16) at
+  the exact M whose goroutine must reach a GC safe point. On the wait4/BSD-stop
+  model there is no thread-correct way to re-deliver it (unlike linux, where a
+  ptrace signal-delivery-stop is per-thread and `PTRACE_CONT(sig)` re-injects to
+  the stopped thread — see the linux note below). Both darwin options have a
+  cost; `resumeAfterSignal` picks the one that keeps the common path reliable —
+  resume with signal 0, which DROPS the intercepted signal (XNU already cleared
+  the thread's pending bit to take the trace-stop; `PT_CONTINUE(0)` does not
+  re-post it):
+  - Re-inject via `ptrace(PT_CONTINUE, pid, 1, SIGURG)`: XNU re-adds the signal to
+    the stopped thread AND calls `psignal()`, which is PROCESS-directed. The target
+    M gets preempted, but the extra process-directed copy lands on an arbitrary
+    thread and ~15% of the time interrupts an M parked in `__psynch_cvwait` → a
+    lost-wakeup wedge (the original churn hang). Fails the E2E gate.
+  - Drop (signal 0, what we do): no spurious copy, so cooperative code — every
+    realistic Go program and the entire E2E suite — is 100% reliable. Cost: a
+    goroutine that is ONLY async-preemptible (a tight loop with no calls,
+    allocations, or channel ops, hence no cooperative safe point) is not preempted
+    for stop-the-world while traced, so `runtime.GC()` in such a target can hang
+    under the debugger. Proven and characterised by `asyncpreempt_e2e_test.go`
+    (env-guarded behind `BINGO_ASYNCPREEMPT_TEST`, not part of the gate).
+
+  Thread-directed re-delivery (reach exactly the intended M, no spurious copy) is
+  impossible under wait4: `PT_THUPDATE` sets the per-thread siglist bit but has no
+  `wakeup(&t->sigwait)`/`task_resume` to drive the stop's resume, and
+  `PT_CONTINUE(sig)` is process-directed. A complete fix needs the Mach-exception
+  stop model (where `SIGURG` is never intercepted), which is out of scope here.
+  Delve takes the same "resume with 0" stance on its wait4 path (`ptraceCont`
+  always passes 0).
 - `PT_STEP` is per-PROCESS on Darwin (despite the API taking a tid) and arms the
   task's *first* thread, not the tid you pass. So an arbitrary thread cannot be
   single-stepped; `needsTempBPStepOver()` returns true and the engine steps over
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -195,21 +195,72 @@ engine advances PC by `len(archTrapInstruction())` and resumes. See the
 ### Darwin / arm64 ([backend_darwin_arm64.go](internal/debugger/backend_darwin_arm64.go))
 
 - Uses ptrace for control flow (`PT_CONTINUE`, `PT_STEP`, `PT_ATTACH`) and
-  Mach (`thread_get_state`, `mach_vm_read/write`, `task_threads`) for state.
-  `PT_ATTACHEXC` is NOT used — it routes signals through Mach exceptions,
-  incompatible with our `wait4`-based loop.
+  Mach (`thread_get_state`, `mach_vm_read/write`, `task_threads`,
+  `thread_set_state`) for state. `PT_ATTACHEXC` is NOT used — it routes signals
+  through Mach exceptions, incompatible with our `wait4`-based loop.
 - `task_for_pid` requires the `com.apple.security.cs.debugger` entitlement
   (the build embeds [entitlements.plist](entitlements.plist) via codesign).
-- `task_threads` returns threads in **creation order**. `threads[0]` is
-  often an idle Go runtime M parked in `pthread_cond_wait`, **not** the
-  goroutine running user code. Darwin `Wait` returns raw SIGTRAP stops without
-  reading Mach thread/register state; the engine resolves the thread sitting at
-  a BRK on its serialized event loop. For SingleStep, save the thread port we
-  issued the step against (`b.stepTID`).
-- `PT_STEP` is per-PROCESS on Darwin (despite the API taking a tid). Always
-  pass `b.pid`, not the Mach thread port.
-- `SIGURG` (Go preemption) and `SIGWINCH` must be re-delivered transparently
-  during both step and continue, or scheduling breaks.
+- **Breakpoints are HARDWARE, not software.** The E2E target is an ad-hoc-signed
+  Go binary; patching its `__TEXT` with a `BRK` invalidates the code signature,
+  and on the next page-in AMFI SIGKILLs the tracee ("has no CMS blob? …
+  Unrecoverable CT signature issue, bailing out"). Instead the backend arms the
+  AArch64 debug registers (`DBGBVR`/`DBGBCR` via `ARM_DEBUG_STATE64`) — see
+  `bingo_set_thread_hw_breakpoints`. No memory is patched, so the signature
+  stays valid and there is no i-cache coherency problem. `WriteMemory`
+  transparently maps the engine's breakpoint pokes onto debug registers: a write
+  of the trap instruction → `armHWBreakpoint`; a write of the original bytes at a
+  known bp → `disarmHWBreakpoint`; anything else → a genuine `rawWriteMemory`
+  (never hit for breakpoints in normal operation).
+- **HW breakpoints are per-thread; re-arm on every resume.** ARM debug registers
+  live in each thread's context, so a thread created after the last arm (Go
+  spins up new M's constantly under load) would not trap. `applyDebugState`
+  re-writes the full breakpoint set into *every* current thread's debug
+  registers on each `ContinueProcess` and each `resumeAfterSignal`. `Wait`
+  resolves a SIGTRAP by scanning threads for one whose PC equals an armed
+  address (`threadAtHWBreakpoint`) and returns `StopBreakpoint{TID, PC}`, so the
+  engine never falls back to a (now nonexistent) in-memory trap scan.
+- **The mid-run re-arm timer must NOT Mach-suspend the task.** Arming on resume
+  covers threads that exist at resume time, but the target can create a new M
+  and migrate the main goroutine onto it *between* two stops (no stop of its own
+  is produced), so `rearmWhileRunning` re-applies the breakpoint set on a
+  `rearmInterval` (10 ms) timer while `wait4` is blocked. It calls
+  `applyDebugState` (a plain per-thread `thread_set_state`) and deliberately does
+  **not** `task_suspend` the whole task. An earlier version suspended the task
+  every tick to force threads off-core (so the CPU reloads their debug registers
+  for certain); that armed reliably but froze the Go runtime — sysmon, the
+  netpoller, the timer-lock holder — at a 10 ms cadence and would occasionally
+  lose `main`'s `time.Sleep` wakeup, deadlocking the tracee with every M parked,
+  correctly armed, and none reaching the breakpoint. Because arming-on-resume is
+  the reliable path and the timer only needs to catch newly-created M's (which
+  context-switch within microseconds on a yielding target and pick up the debug
+  registers from a plain write), dropping the suspend removes the deadlock
+  without reintroducing missed breakpoints. The timer is skipped while
+  single-stepping.
+- **Do NOT re-inject async signals — resume with signal 0.** This is the crux of
+  correct Go-preemption support. Go async preemption is *thread-directed*: the
+  runtime `pthread_kill`s `SIGURG` (16) at the exact M whose goroutine must reach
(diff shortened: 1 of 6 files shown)


xsachax:
Superseded by #70 (surviving darwin solution). sig=0 drops the intercepted SIGURG (drop horn) and breaks non-cooperative STW-GC preemption; #70 stays on the re-inject horn. Closing in favor of #70.

PR #73, opened 2026-07-01
test(debugger): real-backend E2E acceptance gate for backend fixes

xsachax:
## What this is

The shared **acceptance gate** for the debugger-backend fix effort: a real-backend
end-to-end test harness plus a CI workflow. It contains **no backend changes** — only
the test + workflow (393 insertions, 0 modifications to existing files). Every one of
the 10 candidate-fix PRs is branched off this and targets it, so each fix diff is
reviewed in isolation and CI runs this gate against it.

## Files

- `internal/debugger/e2e_integration_test.go` (build tag `e2e`): drives the **real**
  native backend (ptrace on linux/amd64, ptrace+Mach on darwin/arm64) against two
  embedded targets:
  - `TestE2EBasic` — correctness: `Continue → BreakpointHit` (exact line) then
    `StepOver → Stepped` (must advance past the BP line), N iterations. This is what
    fails on the broken linux backend.
  - `TestE2EChurn` — robustness/race: the same loop against a multi-threaded churn
    target, N iterations, asserting no hang (per-op timeout), error, or unexpected exit.
    This is what surfaces the darwin single-step race.
  Tunables: `BINGO_E2E_ITERS`, `BINGO_E2E_CHURN_ITERS`. (`BINGO_E2E_DEBUG=1` adds event
  logging — do **not** set it during race verification; it perturbs timing and hides the
  darwin race.)
- `.github/workflows/debugger-e2e.yml` (`on: push`, all branches):
  - `linux-e2e` (ubuntu-24.04): `go test -tags e2e -race` — authoritative for linux.
  - `darwin-e2e` (macos-14): codesigns the test binary with `entitlements.plist` and
    runs it; `continue-on-error: true` because GH runners don't guarantee `task_for_pid`.
    Darwin is verified authoritatively **locally** (Apple Silicon, codesigned).

## Baseline is intentionally RED

On the current `57-refactor-debugger-foundation` baseline, `linux-e2e` **fails** at the
single-step-past-breakpoint path — that red run *is* the reproduction of the linux bug.
Merging this harness alone leaves CI red until a backend fix lands; each candidate-fix PR
turns it green. Recommend merging the harness together with (or just ahead of) the chosen
fix.

## Why a separate PR

So the verification infrastructure can be reviewed and merged independently of whichever
backend fix(es) you choose, and so the gate is identical across all 10 fix PRs.

Draft until you pick the fix(es) to pair it with.


xsachax:
Removing the E2E harness + CI gate for now (commit 04e3acf) — they were added to this base branch without prior discussion and lived in the wrong place (`internal/debugger/` rather than `test/integration/`). `debugger-e2e-base` now matches `57-refactor-debugger-foundation` with no delta. Will re-add later as proper platform-split Ginkgo integration tests under `test/integration/`. Exact copies of both files are stashed for the redo.

PR #74, opened 2026-07-02
Darwin fix: isolate single-step to the trapped thread (suspend siblings + suspended spawn)

xsachax:
## Root cause

Two darwin/arm64 backend defects:

1. **Wrong-thread single-step (the deterministic step-over hang).** Darwin's `PT_STEP` sets the AArch64 single-step flag (`MDSCR_EL1.SS`) on `get_firstthread(task)` — **not** the trapped thread. When step-over single-steps off a breakpoint, the wrong thread advances one instruction while the trapped thread does not, so the next resume wedges in `wait4`.
2. **Racy attach.** Launch used `PT_ATTACH` on an already-running image, racing `CS_KILL`/`CS_DEBUGGED` codesign state on Apple Silicon and intermittently getting the target killed.

## Fix

- **`isolateForStep`**: Mach-suspend every task thread except the target, arm `MDSCR_EL1.SS` on the target's own thread port, single-step, then `endStepIsolation` resumes the siblings (LLDB/debugserver-style isolation). `get_firstthread` is always among the suspended set, so the stray per-process single-step bit can never advance an unintended thread.
- **Suspended spawn**: `posix_spawn` the target SUSPENDED, then `PT_ATTACH` on the post-exec image (clearing `CS_KILL|CS_HARD`, setting `CS_DEBUGGED`) before `task_resume`, for a deterministic attach.
- Drop a caught SIGURG (resume with signal 0) rather than re-inject it to the wrong thread.

## Verification (coordinator)

- `TestE2EBasic` (iters=25): **8/8 PASS**, 0 hang, 0 data race.

## Known limitation — the wait4 async-preemption trilemma

This fix is on the **"drop SIGURG / suspend-siblings"** horn of a proven trilemma under the `wait4`/BSD-ptrace stop model. Because the intercepted async-preemption SIGURG is dropped (and siblings are suspended during the step), a **purely async-preemptible** goroutine — one with *no* cooperative safepoints, e.g. a call-free `for { x++ }` loop — is **not** stop-the-world preempted while a breakpoint is being serviced. A coordinator GC-preempt probe (non-cooperative spinner + STW `runtime.GC()`) hangs at the GC under this backend.

This is the **same tradeoff class as `GODEBUG=asyncpreemptoff`**. The alternative horn (re-injecting SIGURG, as in #71 / #70-default) keeps async preemption alive ~85% of the time but reintroduces a ~15% SIGURG-misdirection condvar lost-wakeup. **No behavior-preserving fix exists under `wait4`**; the only complete fix is switching the stop source to **Mach exception ports** (like Delve), which is out of scope per `AGENTS.md` and logged as future work.

Realistic Go targets (any call/alloc/channel op is a safepoint) and the E2E gate do not hit the pathological corner.

---
*Draft: one of five distinct darwin/arm64 candidate fixes. Coordinator-finalized from the agent's implementation; base is the shared E2E gate branch `debugger-e2e-base`.*

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -207,7 +207,26 @@ engine advances PC by `len(archTrapInstruction())` and resumes. See the
   a BRK on its serialized event loop. For SingleStep, save the thread port we
   issued the step against (`b.stepTID`).
 - `PT_STEP` is per-PROCESS on Darwin (despite the API taking a tid). Always
-  pass `b.pid`, not the Mach thread port.
+  pass `b.pid`, not the Mach thread port. Worse, XNU arms the CPU single-step
+  (`MDSCR_EL1.SS`) on `get_firstthread(task)` — usually a parked runtime M, NOT
+  the thread that hit the breakpoint — so a bare `PT_STEP` lets every thread run
+  and the kernel may retire the step on the wrong thread. The one that hit the
+  BP then runs straight through the byte-restored breakpoint window before the
+  trap is reinstalled, and the next `Continue` wedges: the ~2% single-step hang.
+  **Fix (`isolateForStep`/`endStepIsolation`, mirrors lldb-debugserver / Delve
+  `singleStep`):** before `PT_STEP`, Mach-`thread_suspend` every thread except
+  the target and set the hardware `MDSCR_EL1.SS` bit on the target only; after
+  the step trap retires, clear the SS bit on ALL threads (XNU left a stray one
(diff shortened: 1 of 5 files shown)


xsachax:
Superseded by #70 (surviving darwin solution). Its distinct pieces sit on the drop-SIGURG horn of the async-preemption trilemma; #70 keeps preemption alive via SIGURG re-injection. Closing in favor of #70.

PR #75, opened 2026-07-18
Darwin fix: stable Mach-port lifecycle + trapped-thread single-step

xsachax:
## Root cause

Three darwin/arm64 backend defects:

1. **Mach-port leak / staleness.** `task_for_pid` leaks a task send-right on every call, and thread ports go stale under thread churn — unbounded Mach-port growth and operations issued against dead ports, causing instability.
2. **Wrong-thread single-step.** `PT_STEP` arms `MDSCR_EL1.SS` on `get_firstthread(task)`, not the trapped thread, so step-over from a breakpoint advances the wrong thread and the next resume wedges in `wait4`.
3. **SIGURG misdirection.** A caught SIGURG re-injected via ptrace is delivered to `get_firstthread(task)`, preempting the wrong M and corrupting the Go runtime's async preemption.

## Fix

- **Port lifecycle**: cache the single task port; track and deallocate thread ports so the backend stops leaking send-rights and never operates on a stale port.
- **Trapped-thread single-step**: arm `MDSCR_EL1.SS` on the trapped thread's own Mach port (LLDB/debugserver-style) instead of relying on per-process `PT_STEP`.
- Drop a caught SIGURG (resume with signal 0) rather than mis-target it; the Go runtime re-issues preemption hints.

## Verification (coordinator)

- `TestE2EBasic` (iters=25): **8/8 PASS**, 0 hang, 0 data race.
- `TestE2EChurn`: green.

## Known limitation — the wait4 async-preemption trilemma

This fix is on the **"drop SIGURG"** horn of a proven trilemma under the `wait4`/BSD-ptrace stop model. Because the intercepted async-preemption SIGURG is dropped, a **purely async-preemptible** goroutine (no cooperative safepoints, e.g. a call-free `for { x++ }` loop) is **not** stop-the-world preempted while a breakpoint is serviced. A coordinator GC-preempt probe (non-cooperative spinner + STW `runtime.GC()`) hangs at the GC under this backend — the **same tradeoff class as `GODEBUG=asyncpreemptoff`**.

The alternative horn (re-injecting SIGURG, as in #71 / #70-default) keeps async preemption alive ~85% of the time but reintroduces a ~15% SIGURG-misdirection condvar lost-wakeup. **No behavior-preserving fix exists under `wait4`**; the complete fix is switching the stop source to **Mach exception ports** (like Delve), out of scope per `AGENTS.md` (future work).

The **port-leak fix (defect 1) is independently valuable** and correct regardless of the SIGURG policy chosen.

---
*Draft: one of five distinct darwin/arm64 candidate fixes. Coordinator-finalized from the agent's implementation; base is the shared E2E gate branch `debugger-e2e-base`.*

### DIFF
--- a/internal/debugger/backend_darwin_arm64.go
+++ b/internal/debugger/backend_darwin_arm64.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"unsafe"
 )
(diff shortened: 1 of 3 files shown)


xsachax:
Superseded by #70 (surviving darwin solution). Its Tier-1 contribution — per-thread Mach send-right deallocation — has been ported into #70 (commit a9b8a07). The remaining changes sit on the drop-SIGURG horn of the async-preemption trilemma, which #70 deliberately avoids.

PR #84, opened 2026-07-19
fix(debugger): add fallback backend for unsupported platforms

### LINKED ISSUE #61: Remove support for unused OS-ARCH combos
99.999% of people who would use this are either on

- Linux x86_64 (let's be real, debugger is not running on x86 32 bit)
- Mac Silicon (darwin arm64, nobody uses intel chips anymore)
- Windows but they use WSL, so it's linux

So it's only worth supporting
- Linux amd64
- Darwin Arm64

This will greatly reduce our scope. Let's mark other versions as unsupported.

xsachax:
## Summary

Closes #61

Building bingo for any GOOS/GOARCH other than `linux/amd64` or
`darwin/arm64` (with `-tags bingonative`) previously failed with a cryptic
`undefined: newBackend` compiler error, since no `backend_*.go` file
provided the platform-specific hooks (`newBackend`, `startTracedProcess`,
`attachToProcess`, `killProcess`).

## Changes

- Added `internal/debugger/backend_unsupported.go`, guarded by a
  `//go:build` constraint matching every unsupported combination
  (including `darwin/arm64` built *without* the `bingonative` tag). It
  builds cleanly everywhere and instead panics with a clear,
  actionable "unsupported platform" message — naming the actual
  supported platforms — as soon as `debugger.New()` is called.
- Updated `README.md` and `AGENTS.md` to document the new fallback
  behavior.

## Verification

- `just build linux amd64` — succeeds
- `just build darwin arm64` — succeeds (codesigned)
- `GOOS=windows GOARCH=amd64 go build ./cmd/bingo` — now compiles
  successfully (previously failed with `undefined: newBackend`);
  confirmed the runtime panic message is clear via a temporary test
  invoking `debugger.New()` under an unsupported build tag combo
- `go test -tags bingonative ./...` — all packages pass
- `go test ./...` (no bingonative tag, darwin) — now also compiles and
  passes, using `fakeBackend` in unit tests as before

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,7 +18,11 @@ Built and tested only on:
   `com.apple.security.cs.debugger` entitlement (or SIP off)
 - `linux/amd64`
 
-Other GOOS/GOARCH combos fail with `undefined: newBackend`.
+Other GOOS/GOARCH combos (and darwin/arm64 without `-tags bingonative`) build
+fine via the fallback backend in
+[backend_unsupported.go](internal/debugger/backend_unsupported.go), but panic
+with a clear "unsupported platform" message the moment `debugger.New()` is
+called.
 
 ## Layout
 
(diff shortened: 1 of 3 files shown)


xsachax:
Closing as won't-do: decided against a runtime-panic fallback backend. Unsupported platforms should fail to build entirely rather than compiling with a runtime panic.

PR #134, opened 2026-08-09
perf(vscode): remove dead keyboard DOM scans

### LINKED ISSUE #129: [vscode] Remove dead DOM scans from tree keyboard navigation
**Describe the bug**

`editors/vscode/src/webviewApp.ts:356-361` computes keyboard-navigation candidates by filtering `session.tree.nodes`, querying `activeScene` once per node, and rejecting elements with the CSS class `filtered`.

No source or generated bundle assigns that class, CSS cannot add it, and search filtering already structurally removes nodes through `filteredSession` -> `filterFullTree` before rendering. Existing DOM tests also assert filtered nodes are absent rather than class-marked. The predicate therefore accepts every node and only adds dead DOM work.

Because the rendered tree is capped at 500 goroutines, each arrow-key event can perform up to 500 needless selector queries before the separate query used to focus the next node. This is bounded redundant work and dead code; it is not a claimed guaranteed O(n^2) regression.

**To Reproduce**

1. Inspect the arrow-key handler in `editors/vscode/src/webviewApp.ts:351-375`.
2. Observe that it calls `activeScene.querySelector()` for every entry in `session.tree.nodes` and checks `classList.contains("filtered")`.
3. Search the VS Code extension source and bundle inputs for code that assigns the `filtered` class; there is none.
4. Inspect `filteredSession` in `editors/vscode/src/webviewApp.ts:509-520` and `filterFullTree` in `editors/vscode/src/tree.ts`; filtered nodes are omitted from `session.tree.nodes` before rendering.
5. Dispatch an arrow key while a large tree is rendered and instrument `activeScene.querySelector()`; the handler performs one query per rendered node even though the candidate list is unchanged.

**Expected behavior**

Keyboard navigation should use `session.tree.nodes` directly and query the DOM only for the selected next node that needs focus.

**Screenshots**

N/A. This is statically reproducible and does not change rendered output.

 - Architecture: All supported architectures
 - Version: `main` at `0509b53` / VS Code extension `0.3.1`

**Additional context**

Priority: **P3**. User-visible behavior remains correct, but every arrow key performs bounded needless DOM scans in a hot interaction path and the dead predicate obscures the structural filtering invariant. A deterministic DOM test can guard selector-call count without relying on timing.

xsachax:
## Summary

- use the already-filtered `session.tree.nodes` directly during arrow-key navigation
- remove up to 500 dead per-node selector calls for a CSS class that is never assigned
- guard the interaction with a deterministic DOM test that permits only the final focus lookup

This does not bump the VS Code extension version: rendered behavior, protocols, and packaging are unchanged, so this is not a material shipped behavior change under the `AGENTS.md` versioning invariant.

## Tests

- `npm --prefix editors/vscode run lint`
- `npm --prefix editors/vscode run typecheck`
- `npm --prefix editors/vscode run test:unit`
- `npm --prefix editors/vscode run test:dom`
- `npm --prefix editors/vscode run build`
- `npm --prefix editors/vscode run package:list`

Fixes #129

### DIFF
--- a/editors/vscode/src/webviewApp.ts
+++ b/editors/vscode/src/webviewApp.ts
@@ -353,12 +353,7 @@ function renderGraph(
       return;
     }
     event.preventDefault();
-    const visible = session.tree.nodes.filter((node) => {
-      const element = activeScene.querySelector<SVGGElement>(
-        `[data-goid="${String(node.goroutine.id)}"]`,
-      );
-      return element?.classList.contains("filtered") !== true;
-    });
+    const visible = session.tree.nodes;
     const current = visible.findIndex(
       (node) => node.goroutine.id === session.selectedGoroutine,
     );
--- a/editors/vscode/test/webviewDom.test.ts
+++ b/editors/vscode/test/webviewDom.test.ts
@@ -98,6 +98,15 @@ describe("concurrency webview DOM", () => {
     })(model());
     const first = document.querySelector<SVGGElement>('[data-goid="1"]')!;
     const second = document.querySelector<SVGGElement>('[data-goid="2"]')!;
+    const scene = document.querySelector<SVGGElement>("svg > g")!;
+    const querySelector = scene.querySelector.bind(scene);
+    let selectorCalls = 0;
+    Object.defineProperty(scene, "querySelector", {
+      value: (selector: string): Element | null => {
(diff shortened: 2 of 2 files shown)


xsachax:

PR #139, opened 2026-08-09
fix(debugger): route emit logs through session logger

### LINKED ISSUE #131: [debugger] engine.emit diagnostics bypass the injected session logger
**Describe the bug**
Priority: P3

`internal/debugger/engine.go` documents `engine.log` as the single sink for engine logging, but `engine.emit` still sends its two rare-path diagnostics through the package-level `slog` functions:

- `internal/debugger/engine.go:1141`: marshal failures use `slog.Error`.
- `internal/debugger/engine.go:1151`: event-buffer overflow uses `slog.Warn`.

The root cause is an incomplete conversion in commit `d316313` (PR #83): that change injected a per-session logger and converted the other engine call sites, but these two calls remained on `slog.Default()`.

As a result, these diagnostics bypass the configured engine logger, its log level/handler, and its `session` correlation attribute. They appear in the global stream instead, violating the engine comment and the repository's per-session logging invariant. This is low-frequency and does not change debugger control flow, so priority is P3.

**To Reproduce**
1. In a same-package `internal/debugger` test, save `slog.Default()` and install a capture handler as the temporary global default.
2. Construct an `engine` with a different capture logger and a capacity-one `events` channel.
3. Call `engine.emit` with an unsupported JSON payload such as `make(chan struct{})` to force `protocol.NewEvent` to return a marshal error.
4. Emit one valid event to fill the channel, then emit a second valid event to force the non-blocking overflow branch.
5. Inspect both capture handlers, then restore the saved global logger.

Observed: both `engine.emit: marshal event failed` and `engine.emit: events buffer full — dropping` are captured only by the global/default handler; the injected engine handler receives neither.

**Expected behavior**
Both diagnostics should be emitted through `e.log` so the injected per-session handler receives them and the global/default handler receives nothing. The fix should only replace the two logging receivers and must not alter event sequencing, buffering, or debugger control/state/native execution behavior.

**Screenshots**
N/A — this is a structured-logging routing defect reproduced with capture handlers.

 - OS: macOS / Darwin 25.5.0
 - Architecture: arm64
 - Version: `origin/main` at `0509b53bfc44da889e152ee20f56ce3663882c65`; Go 1.26.1

**Additional context**
Issue #32 reported duplicate/unhelpful debugger logging and was closed by `d316313` / PR #83. That fix established `e.log` and the explicit invariant against package-level `slog`, but missed these two `engine.emit` branches.

Verification evidence: a controlled same-package test with distinct global and injected capture handlers triggered both paths deterministically and showed both records only on `slog.Default()`. Two independent rubber-duck reviews accepted the causal link and the receiver-only `e.log.Error` / `e.log.Warn` correction as safe. An all-state repository issue/PR search immediately before filing found only the closed historical PRs #83 and #80, with no active duplicate.

xsachax:
## Summary
- route `engine.emit` marshal-failure and event-buffer-overflow diagnostics through the injected per-session logger
- add a deterministic same-package regression test covering both rare paths and guarding against global logger leakage

Fixes #131

## Testing
- `go test -tags bingonative ./internal/debugger -run '^TestEngineEmitUsesInjectedLogger$' -count=1`
- `go test -tags bingonative ./internal/debugger/... -count=1`
- `go vet -tags bingonative ./internal/debugger/...`
- `go test -tags bingonative -race ./internal/debugger/... -count=1`

### DIFF
--- a/internal/debugger/engine.go
+++ b/internal/debugger/engine.go
@@ -1138,7 +1138,7 @@ func (e *engine) nextSeq() uint64 {
 func (e *engine) emit(kind protocol.EventKind, payload any) {
 	evt, err := protocol.NewEvent(kind, e.nextSeq(), payload)
 	if err != nil {
-		slog.Error("engine.emit: marshal event failed", "kind", kind, "err", err)
+		e.log.Error("engine.emit: marshal event failed", "kind", kind, "err", err)
 		return
 	}
 	// Non-blocking on purpose: this runs on the serialized loop, so blocking
@@ -1148,7 +1148,7 @@ func (e *engine) emit(kind protocol.EventKind, payload any) {
 	select {
 	case e.events <- evt:
 	default:
-		slog.Warn("engine.emit: events buffer full — dropping", "kind", kind)
+		e.log.Warn("engine.emit: events buffer full — dropping", "kind", kind)
 	}
 }
 
--- a/internal/debugger/engine_emit_test.go
+++ b/internal/debugger/engine_emit_test.go
@@ -1,0 +1,85 @@
+package debugger
+
+import (
+	"context"
+	"log/slog"
+	"sync"
+	"testing"
+
+	"github.com/bingosuite/bingo/pkg/protocol"
+)
+
+type emitLogRecord struct {
+	level   slog.Level
+	message string
+}
+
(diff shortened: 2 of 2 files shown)


xsachax:

PR #152, opened 2026-08-09
fix(dap): retry transient accept failures

### LINKED ISSUE #151: [server] Retry transient DAP accept failures
**Describe the bug**
Priority: **P2**. A single transient DAP listener `Accept` failure permanently exits the accept loop. The listener remains open, the server retains `dapStarted`/`dapAddress`, `/api/health` continues reporting DAP enabled, and `StartDAP` refuses recovery with `ErrDAPAlreadyStarted`. New TCP handshakes can enter the kernel backlog, but no handler accepts them, so valid DAP clients hang until the whole bingo process is restarted.

Root cause at `0509b53bfc44da889e152ee20f56ce3663882c65`:
- `internal/dap/server.go:acceptLoop` logs every non-close `Accept` error and immediately returns.
- Unlike `net/http.Server.Serve`, it does not retry temporary errors with capped backoff.
- `internal/server/dap.go:StartDAP` publishes the listener once and has no failure callback.
- `internal/server/handler.go:handleHealth` derives `dap.enabled` only from non-empty `dapAddress`, which is cleared only during full server shutdown.

**To Reproduce**
1. Build bingo at commit `0509b53`.
2. Run it with HTTP and DAP on loopback under a small soft fd limit, for example `(ulimit -n 32; exec bingo -addr 127.0.0.1:0 -dap-addr 127.0.0.1:0 -v)`.
3. Open bare TCP connections to the advertised DAP address until the server logs `dap accept error ... too many open files`.
4. Close those connections so file descriptors are available again.
5. Query `/api/health`; it still reports `dap.enabled: true` and the original address.
6. Connect to that address and send a valid DAP `initialize` request. The TCP connect succeeds into the backlog, but no response arrives because the accept loop has exited.

Observed deterministic evidence: 80 connection attempts triggered `accept: too many open files`; after closing them, health remained enabled and a fresh initialize timed out.

**Expected behavior**
Temporary accept failures should be logged and retried with bounded exponential backoff, resetting the delay after a successful accept, as the HTTP listener already does. Shutdown must still unblock the backoff promptly and close all handlers.

**Screenshots**
N/A

 - OS: macOS
 - Architecture: arm64
 - Version: commit `0509b53bfc44da889e152ee20f56ce3663882c65`

**Additional context**
The lowered fd limit makes the system condition deterministic; it is not load-bearing to the defect, since any transient `Accept` error follows the same permanent-exit path. Focused searches across all open/closed issues and PRs found no duplicate.

xsachax:
## Summary
- retry non-shutdown DAP accept failures with capped exponential backoff
- reset the delay after a successful accept
- exit immediately for a closed listener and let shutdown interrupt a pending retry wait
- cover repeated accept failures, explicit listener closure, and shutdown during backoff deterministically

## Validation
- `golangci-lint run --build-tags bingonative ./...`
- `go tool goimports -l .`
- `go test -tags bingonative -race ./cmd/bingo ./internal/server ./internal/dap`
- `go test -tags bingonative -race -count=20 ./internal/dap`
- targeted accept-loop tests under `-race -count=100`
- `go vet -tags bingonative ./cmd/bingo ./internal/server ./internal/dap`

Fixes #151

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1166,6 +1166,12 @@ polling has no lifecycle effect. Positive idle durations below `1ms` or with a
 fractional millisecond are rejected so the timer and integer `timeoutMs` field
 always describe the exact same interval.
 
+**Listener safety.** The CLI and `just server` defaults bind management,
+WebSocket, and DAP endpoints to explicit IPv4 loopback. The protocols do not
+authenticate clients, and WebSocket clients without an `Origin` header are
+valid, so never restore wildcard defaults. Operators may explicitly choose a
+non-loopback address for trusted-network or externally authenticated setups.
+
 **Connect-or-start and ownership.** A frontend health-checks the known
 management address and reuses a compatible process; otherwise it starts bingo
 and lets listener binding arbitrate concurrent startup attempts. Frontends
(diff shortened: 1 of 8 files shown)


xsachax:

PR #155, opened 2026-08-09
test(vscode): clean up packaged process groups

### LINKED ISSUE #153: [vscode] Packaged E2E failure cleanup can orphan launched debuggees
**Describe the bug**

At `origin/main` `0509b53bfc44da889e152ee20f56ce3663882c65`, failure cleanup in `editors/vscode/test/packagedE2E.ts:169-177` sends `SIGKILL` only to `child.pid`. The server is spawned with `detached: true` in `editors/vscode/src/serverProcess.ts:53-57`, so it leads a test-owned process group, and debuggees launched through DAP inherit that group.

There is a live-debuggee failure window from `runExample`'s launch (`packagedE2E.ts:201`) through multiple timed waits/assertions until disconnect (`packagedE2E.ts:330-335`). A failure in that window bypasses bingo's graceful signal -> `Server.Shutdown` -> hub -> debugger `Kill` path. Exact-PID `SIGKILL` removes only bingo and can leave the inherited debuggee alive and reparented. On Darwin, that tracee can remain Mach-suspended indefinitely on persistent/self-hosted runners. Linux often self-heals, and hosted runners are ephemeral, but the cleanup is still non-hermetic.

The current liveness guard is also unsafe: Node keeps `child.exitCode === null` after signal termination and reports the signal in `child.signalCode`, so PID reuse could make a later group signal dangerous unless cleanup is gated by an explicit exit/error latch.

**To Reproduce**

1. Spawn a Node parent with `{ detached: true }` so its PID is the process-group ID.
2. Have that parent spawn a long-lived child that inherits the group.
3. Send `SIGKILL` to the positive captured parent PID and await the parent exit.
4. Probe the captured child PID and negative process-group ID with signal `0`; both remain alive.
5. Send `SIGKILL` to the captured negative process-group ID and probe until `ESRCH`; the group disappears.

Hermetic local Darwin reproduction (captured PIDs only) observed:

```json
{"parentPid":74730,"childPid":74731,"childAliveAfterPidKill":true,"groupAliveAfterPidKill":true,"groupGone":true}
```

No name-based signaling is involved.

**Expected behavior**

Packaged E2E failure cleanup should first send graceful `SIGTERM` to the exact captured server PID so bingo can resume, kill, and reap Darwin tracees. If the explicitly tracked server has not exited within a bounded grace, the test may escalate to `SIGKILL` for only the captured owned process group. Cleanup must await the server outcome, confirm the process group is absent with a bounded `kill(-pgid, 0)`/`ESRCH` check, aggregate cleanup errors with the original failure, and preserve scratch/log evidence when cleanup is incomplete. The normal success path must continue waiting for server-owned idle exit, and production extension management must remain never-kill.

**Screenshots**

N/A

 - OS: Darwin (highest impact); Linux is also non-hermetic
 - Architecture: arm64 on Darwin; amd64 on Linux
 - Version: `0509b53bfc44da889e152ee20f56ce3663882c65`

**Additional context**

`editors/vscode/scripts/owned-process.mjs` already has test-only group cleanup for the packaged smoke test, but it immediately group-SIGKILLs and only awaits the server exit; it does not prove the owned group is gone. The repair should be shared/DRY if safe and covered by deterministic fast tests for descendant removal, already-exited no-signal behavior, timeout escalation, and the absence of bare `process.kill(child.pid, "SIGKILL")` in packaged E2E.

`AGENTS.md:938-941` deliberately says the packaged E2E may signal only the exact server PID. That invariant contradicts the inherited-debuggee ownership model and must be corrected alongside the test-only cleanup. No production server-kill behavior should be added.

xsachax:
## Summary

- track detached packaged servers with an explicit exit/error latch and one-way owned-group state before cleanup signaling
- request graceful exact-PID shutdown, then bound and verify owned process-group escalation even when the server leader exits before descendants
- aggregate cleanup failures, preserve scratch evidence, and align the documented test-only ownership invariant

## Testing

- `npm --prefix editors/vscode run check` (119/119 tests)
- cleanup process tests repeated 25/25 runs
- `just vscode-package`
- `npm --prefix editors/vscode run smoke:server`
- `just build-examples && npm --prefix editors/vscode run e2e:packaged`

Fixes #153

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -938,11 +938,14 @@ requires local/self-hosted Apple Silicon, where the same E2E covers all five
 examples. Both tests observe server-owned idle exit on success. On failure the
 test harness tracks the detached child's terminal outcome before signaling,
 requests graceful shutdown with SIGTERM to the exact captured server PID, and
-escalates to SIGKILL on only its captured process group when the tracked server
-remains live past a bounded grace. It then waits for the server outcome and
-confirms the owned group is absent before deleting scratch artifacts; incomplete
-cleanup is reported and preserves those artifacts. Cleanup is test-only and
-must never enter extension production code.
+escalates to SIGKILL on only its captured process group when that owned group
+remains nonempty past a bounded grace, even if the server leader exited first.
+The tracker permanently releases group ownership when an existence probe reports
+it absent, so cleanup never signals a later group that reused the numeric PGID.
+It then waits for the server outcome and confirms the owned group is absent
+before deleting scratch artifacts; incomplete cleanup is reported and preserves
+those artifacts. Cleanup is test-only and must never enter extension production
+code.
 Apple's external linker can vary `LC_UUID` even for identical cgo inputs, but
 current dyld rejects binaries with `-no_uuid`; `normalize-mach-o-uuid.mjs`
 therefore derives a stable UUID from the unsigned Mach-O with its UUID and
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -936,9 +936,13 @@ activation/view/custom-event acknowledgement test; linux additionally runs the
 real packaged DAP→WebSocket graphical-model E2E. Darwin native-debug execution
 requires local/self-hosted Apple Silicon, where the same E2E covers all five
 examples. Both tests observe server-owned idle exit on success. On failure the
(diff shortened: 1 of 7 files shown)


xsachax:

PR #163, opened 2026-08-09
fix(dap): reconcile joined session lifecycle

### LINKED ISSUE #159: [dap] Joined session can miss stop or exit during hub state transition
**Describe the bug**
A DAP client joining an existing managed session can miss the session's stop or exit when admission lands between the debugger event broadcast and the hub's corresponding state transition.

The hub currently broadcasts a suspending or `ProcessExited` event, releases the outbound mutex, and only then calls `transitionState`. PR #141 serializes each outbound frame with admission, so the race is now deterministic when the join lands in that remaining gap, but it does not make the event plus state transition one atomic unit.

Exact stop interleaving:

1. The managed session is `running`.
2. The Run loop broadcasts `EventBreakpointHit` and releases `outboundMu`.
3. A DAP join acquires `outboundMu`, registers after the event fan-out, and receives a welcome `EventSessionState(running)` because the state transition has not happened yet.
4. DAP consumes that first state by clearing `awaitingWelcome`; it received no breakpoint event, so it emits no `stopped` event.
5. The Run loop transitions to `suspended` and broadcasts `EventSessionState(suspended)`.
6. DAP ignores that correction because it only handles the first welcome state.

The identical exit variant broadcasts `EventProcessExited`, admits the joiner with a stale `running` welcome, then broadcasts `EventSessionState(exited)`. The IDE receives neither `exited` nor `terminated` and can remain attached to a dead session.

Priority: **P3**. This requires a manual DAP join during the narrow stop/exit transition window, so likelihood is low. The stop impact is a stuck/running IDE view until a later stop happens; the exit impact is permanent for that connection because no later lifecycle event repairs it.

**To Reproduce**
1. Start a managed session and keep it running.
2. Arrange for `AddClient` to run after a suspending event or `ProcessExited` has been delivered but before `transitionState` updates the managed state.
3. Join that session through DAP (`attach` with `session` and no `pid`).
4. Observe the joiner's ordered state sequence as `[running, suspended]` or `[running, exited]` without the preceding debugger event.
5. Observe that the adapter emits neither the required `stopped(pause)` nor `terminated` event.

This is reproducible deterministically with a hub test hook at the event/transition boundary and with direct DAP state-sequence tests.

**Expected behavior**
A joining client must fall on exactly one side of a stop/exit transition: either it is registered before the debugger event and receives the event followed by `SessionState`, or it is admitted after the transition and receives an authoritative `suspended`/`exited` welcome. It must never miss both sources of lifecycle truth.

DAP joiners should also reconcile later `SessionState` frames idempotently: emit one `stopped(pause)` when state changes to suspended without a stop event, emit one `terminated` for exited, and avoid duplicates when the actual debugger event already updated adapter state.

**Screenshots**
N/A.

 - OS: macOS
 - Architecture: arm64
 - Version: PR #141 head `8a6171c4939b90079425fc652ebd35d862a2a494` (stacked on `main` at `0509b53bfc44da889e152ee20f56ce3663882c65`)

**Additional context**
This is not a duplicate of #132 or PR #141. #132 covers gaps and reordering caused by unicast join-state sequence allocation. #141 fixes that by serializing admission, sequence allocation, and each broadcast under `outboundMu`. The remaining bug is that a suspending/exiting debugger event and its `SessionState` transition are still two separately locked operations, and DAP currently treats only the first join-state frame as authoritative.

The repair should preserve wire order (`BreakpointHit`/`Stepped`/`Paused`/`Panic` or `ProcessExited` first, then `SessionState`) while holding the outbound admission mutex across both. No debugger-internal or wire-protocol change is required.

xsachax:
## Stack

Layer 2 atop #141. Review and merge #141 first; this PR intentionally targets `xsachax-fix-hub-sequence-stream`, not `main`.

## Summary

- hold `outboundMu` across suspending/exiting event fan-out, the state write, and the following `EventSessionState`, preserving event-then-state wire order
- cover `ProcessExited` both in the outer hub loop and while the hub is waiting for a suspended-session resume
- keep joined DAP handlers reconciled with later session states without duplicate `stopped`/`terminated` events, including a fast Continue/Step between a stop and its paired state frame
- begin a new joined-adapter lifecycle on either a successful `EventRestarted` or an observed `exited`/`idle`→`running` fresh launch, including launches driven by another client
- preserve duplicate state suppression within one lifecycle while restoring stop and termination reconciliation for each new process
- document the lifecycle/admission and process-epoch invariants in `AGENTS.md`

## Testing

- `go test -tags bingonative -race -count=1 ./internal/hub/... ./internal/dap/... ./internal/server/...`
- `go vet -tags bingonative ./internal/hub/... ./internal/dap/... ./internal/server/...`
- focused hub transition races repeated 10 times with Ginkgo (`100` admission races per run)
- focused DAP join-state regressions repeated 20 times under `-race`
- owner, requesting-joiner, and passive-joiner restart/exit lifecycle regressions repeated 20 times under `-race`
- joined observer exit→idle→fresh running→stop→exit lifecycle repeated 20 times under `-race`

Fixes #159

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -695,11 +695,21 @@ of issue #32). Constructors accept a nil logger and fall back to
 
 ## Hub seq stream — why one counter
 
-The hub re-stamps every outbound event with its own atomic `seq` counter. The
-engine has its own seq, and the hub also synthesises events (errors,
-confirmations like `BreakpointSet`). If clients saw both streams interleaved,
-they'd see two overlapping monotonic sequences and couldn't detect drops.
-**Always go through `h.seq.Add(1)` before broadcasting.**
+The hub re-stamps every outbound event with its own `seq` counter. The engine
+has its own seq, and the hub also synthesises events (errors, confirmations like
+`BreakpointSet`). If clients saw both streams interleaved, they'd see two
+overlapping monotonic sequences and couldn't detect drops.
+
+**`broadcast` is the only sequence-allocation path.** Its `outboundMu` covers
+sequence assignment, marshal, and nonblocking `deliver` to the registry
+snapshot, so every connected client sees the same frames in the same order.
+Managed `AddClient` takes that same lock across registry admission and a
+**broadcast** `EventSessionState`; existing clients therefore learn the new
+client count without a phantom gap, and the join state is the new client's
+first frame. Do not unicast an allocated sequence, reuse `seq` without
+incrementing it, or route join state through `Run` (which may be blocked in the
+suspended wait). `shutdown` also takes `outboundMu` before registry teardown so
+admission plus its initial state enqueue cannot interleave with closure.
 
 ## Restart — hub-level, not engine-level
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -708,8 +708,21 @@ Managed `AddClient` takes that same lock across registry admission and a
 client count without a phantom gap, and the join state is the new client's
 first frame. Do not unicast an allocated sequence, reuse `seq` without
 incrementing it, or route join state through `Run` (which may be blocked in the
-suspended wait). `shutdown` also takes `outboundMu` before registry teardown so
-admission plus its initial state enqueue cannot interleave with closure.
+suspended wait).
+
+Lifecycle events and their state transitions are one outbound critical section:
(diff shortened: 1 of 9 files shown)


xsachax:
Superseded by merged #141. Hub admission plus stop/exit state delivery are now atomic, closing issue #159 without the obsolete DAP termination-latch layers in this stack.

PR #167, opened 2026-08-09
fix(debugger): expand sibling pointer aliases

### LINKED ISSUE #165: [debugger] Pointer aliases lose eager children after first expansion
**Describe the bug**
`formatCtx.visited` is shared across an entire eager variable root and `formatPointer` never removes a pointee after formatting it. As a result, the first pointer to an address expands, but any later sibling alias to the same non-cyclic target is treated as a cycle and gets no `Children`.

This is order-dependent alias deduplication rather than the documented cycle guard. In DAP, a variable without `Children` receives no `variablesReference`, so the later alias cannot be expanded at all.

Priority: P3.

**To Reproduce**
1. At `origin/main` commit `0509b53`, construct a pure formatter value with type `struct { Left, Right *int }`.
2. Store `0x2000` in both pointer fields and store integer `42` at `0x2000`.
3. Call `dwarfReader.formatTyped` on the struct.
4. Observe deterministic child counts: `Left` has 1 child and `Right` has 0 children.

A focused unit repro reports:

```text
Right children = 0, want 1
```

**Expected behavior**
Both aliases should independently contain the bounded eager pointee child. The visited-address guard should track only the active recursion path, removing an address after that pointer subtree returns, while retaining the existing depth, pointer-dereference, node, and byte limits.

**Screenshots**
N/A.

 - OS: All supported platforms (pure formatter bug; reproduced on macOS)
 - Architecture: All supported architectures (reproduced on darwin/arm64)
 - Version: `origin/main` at `0509b53`

**Additional context**
PR #119 and `AGENTS.md` describe `visited` as a cycle guard and the protocol/DAP tree as eagerly expandable. Global alias suppression is undocumented and leaks into DAP because `buildVarTree` allocates a `variablesReference` only when `Children` is non-empty. Existing `maxValueDepth`, `maxPtrDeref`, and global node/byte budgets continue to bound recursion and repetition after making the guard path-local.

xsachax:
## Summary

- treat pointer-cycle tracking as an active recursion path instead of root-wide alias deduplication
- preserve eager children for sibling aliases while retaining the existing depth, dereference, node, and byte bounds
- cover sibling aliases, path ownership, self/mutual cycles, and unreadable pointees; document the invariant in AGENTS.md

## Validation

- `go test -tags bingonative -race ./internal/debugger ./internal/dap`
- `go vet -tags bingonative ./internal/debugger ./internal/dap`

Fixes #165

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -539,8 +539,10 @@ are detected by a `mach_msg` receive loop.
   `interface {}` / `error` self-reference.
 - **Bounds & fallback (never error the stop):** `maxValueDepth=4`,
   `maxChildren=100` (overflow appended as a synthetic `… N more` node),
-  pointer-deref depth `1`, `maxStringBytes≈256`, and a **visited-address cycle
-  guard** (pointer targets) so self-referential structures can't infinite-loop.
+  pointer-deref depth `1`, `maxStringBytes≈256`, and an **active recursion-path
+  cycle guard** for pointer targets. Pointer addresses are removed as each
+  subtree unwinds, so self-referential structures terminate without suppressing
+  later sibling aliases to the same non-cyclic target.
   Those are *per-path* caps; on top of them a **shared global ceiling**
   (`maxTotalNodes=10000`, `maxTotalBytes=256 KiB`) is threaded through `formatCtx`
   and debited once per `formatNode` and by every read. Without it a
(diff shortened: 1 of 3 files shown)


xsachax:
Superseded by #190, which preserved this exact pointer-cycle commit as the first layer and merged it into main after full race, vet, Linux, and native Darwin validation.

PR #171, opened 2026-08-09
fix(vscode): preserve webview focus on rerender

### LINKED ISSUE #169: [vscode] Preserve webview focus across telemetry rerenders
**Describe the bug**
Every model revision in `editors/vscode/src/webviewApp.ts:24-52` records `document.activeElement`, replaces the entire `#app` subtree with `root.replaceChildren()`, and tries to restore focus. The identity helper at `editors/vscode/src/webviewApp.ts:699-715` recognizes only static element IDs and goroutine `data-goid` values.

The session selector and all toolbar controls lack either identity: the session `<select>` (`webviewApp.ts:80-95`), Refresh and Copy snapshot (`webviewApp.ts:146-160`), and Fit/Zoom out/Zoom in (`webviewApp.ts:233-245`). A selector arrow change posts `selectSession`, the registry publishes a new revision, and the rerender detaches the focused selector without restoring its replacement. This breaks continued keyboard traversal through sessions. The same focus loss affects the buttons, with lower P3 impact.

The current restoration also has the opposite failure: when focus has moved from the webview to the editor, `document.activeElement` can remain the previously focused webview control while `document.hasFocus()` is false. A telemetry revision still calls `.focus()` on the replacement and can yank focus back from the editor.

**To Reproduce**
1. Start at least two bingo debug sessions so the Bingo Concurrency session selector is present.
2. Focus the session selector with the keyboard and press an arrow key to choose another session.
3. Observe the `change` event publish `selectSession`; the registry update renders a new revision and `root.replaceChildren()` removes the focused selector.
4. Press another arrow key. Focus is no longer on the replacement selector, so keyboard traversal cannot continue.
5. Repeat with Refresh, Copy snapshot, Fit, Zoom out, or Zoom in; a telemetry rerender removes the focused button and does not restore its replacement.
6. Focus the search field, tree viewport/node, or another restorable control, then move focus back to the editor and allow a telemetry revision to render. The stale webview `activeElement` may be restored even though the webview document is blurred, pulling focus away from the editor.

**Expected behavior**
Stable controls should have bounded literal focus identities and a rerender should focus the matching new element only when the webview document had focus immediately before replacement. Existing search, viewport, and tree-node restoration should remain intact. If the corresponding control is absent in the new render, restoration should do nothing. A native open `<select>` popup may still close because the full-root renderer rebuilds the control; preserving that popup would require a separate incremental-rendering change.

**Screenshots**
N/A. The problem is reproducible through keyboard focus and DOM identity.

 - Architecture: darwin/arm64 and linux/amd64 packages
 - Version: VS Code extension `0.3.1`, PR #134 head `d0a847e3c0d97aa4f676c17604defe639d5620af`

**Additional context**
Priority: **P2** because the session selector becomes a keyboard trap after each selection-triggered rerender and the blurred-document path can steal focus from the editor. Toolbar button focus loss is a P3 manifestation of the same root cause.

This is a separate layer above draft PR #134. That PR removes dead selector scans from tree arrow navigation but intentionally does not change rendered behavior or bump the extension version. The focus fix changes shipped webview behavior and should bump the extension package to `0.3.2`.

xsachax:
## Stack

**Layer 2 atop #134.** This PR intentionally targets `xsachax-fix-vscode-keyboard-navigation`; #134 remains the lower performance-only layer.

## Summary

- give the session selector and Refresh, Copy, Fit, Zoom out, and Zoom in controls bounded literal focus identities
- preserve those controls plus the existing search, viewport, and goroutine-node focus across full-root model rerenders
- gate restoration on the webview document having focus before replacement so telemetry cannot pull focus back from the editor
- treat a missing replacement control as a no-op and retain the full-root renderer; an open native select popup may still close on rebuild
- replace linkedom's inert focus behavior with a prototype-level `activeElement`/`hasFocus` shim and cover selector traversal, every button, existing identities, blur, and absent targets
- bump the shipped VS Code extension boundary from 0.3.1 to 0.3.2 and update the manifest/package verifier and agent invariant

## Validation

- `npm --prefix editors/vscode run check`
- `npm --prefix editors/vscode run test:integration`
- `npm --prefix editors/vscode run package:reproducible`
- `npm --prefix editors/vscode run package:verify`
- `npm --prefix editors/vscode run smoke:server`

Fixes #169

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -919,7 +919,7 @@ target metadata, architecture, mode, and entitlements.
 The extension package version is the installed-runtime upgrade boundary:
 material shipped behavior changes must bump both `package.json` and the lockfile
 or VS Code can retain an older bundle under the same identity. The manifest test
-and package verifier pin the current version (**0.3.1**) in source and VSIX
+and package verifier pin the current version (**0.3.2**) in source and VSIX
 metadata.
 The root Run and Debug dropdown exposes exactly two `"type":"bingo"` choices:
 launch one of five progressive examples through a `pickString`, and join a
@@ -1010,7 +1010,14 @@ the 500-node rendering cap, re-lays out each match with at most four ancestors,
 and resets fit so a deep or previously capped match cannot remain invisible.
 Empty results keep Fit/zoom callbacks safe even without an SVG scene. SVG
 treeitems carry `aria-level`, sibling position/size, selection, and parent
-context; arrow navigation moves DOM focus with selection.
+context; arrow navigation moves DOM focus with selection. Full-root model
+rerenders preserve focus through bounded literal identities for the session
+selector and controls, the two static element IDs, and numeric goroutine IDs,
+but only when `document.hasFocus()` was true before replacement; a blurred
+webview must never reclaim focus from the editor, and an identity absent from
+the new render is a no-op. Rebuilding the selector can still close its native
+open popup; preserving that transient browser UI would require an incremental
+renderer and is intentionally outside this invariant.
 
 ### Handshake (Delve-style, VS Code-compatible)
 
--- a/editors/vscode/README.md
+++ b/editors/vscode/README.md
@@ -20,8 +20,9 @@ just vscode-install
 ```
 
 This builds, verifies, and installs `dist/bingo-<platform>.vsix`. The graphical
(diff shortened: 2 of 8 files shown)


xsachax:

Issue #128, opened 2026-08-09
[dapcli] clear 0 deletes a pending breakpoint


xsachax:
**Describe the bug**
`cmd/dapcli` stores buffered breakpoints with the zero-value `id == 0` until a `setBreakpoints` response assigns the debugger's real positive ID (`cmd/dapcli/main.go:44-48`, `:463-480`, `:539-546`). `clearBreakpoint` accepts any integer and matches solely on `bp.id == id` (`cmd/dapcli/main.go:483-509`), so `clear 0` removes the first pending breakpoint it encounters. With pending entries in multiple files, map iteration also makes the victim nondeterministic. If configuration is already active while IDs are still pending, the client can additionally send an unintended replace-all `setBreakpoints` diff for the affected file.

**To Reproduce**
1. Start `cmd/dapcli` and connect it to a bingo DAP listener, but do not launch or attach yet.
2. Enter `break first.go:10`.
3. Enter `break first.go:20`.
4. Enter `clear 0`.
5. Observe `breakpoint 0 cleared`; inspecting `bpsByFile["first.go"]` shows the first buffered breakpoint was deleted even though real debugger breakpoint IDs start at 1.

This is deterministic with two pending entries in one file because the slice scan always matches the first zero-valued ID. The same bug can occur during the configured/in-flight window before a `setBreakpoints` confirmation populates IDs, where clearing zero also triggers a replacement request.

**Expected behavior**
Non-positive IDs are not valid debugger breakpoint IDs. `clear 0` (and negative IDs) should report `no breakpoint with id 0`, leave every pending breakpoint unchanged, send no DAP request, and never print a successful clear message. Clearing a valid positive ID should continue to remove that breakpoint and synchronize the file's replacement set normally.

**Screenshots**
N/A.

**Desktop (please complete the following information):**
 - OS: macOS
 - Architecture: arm64
 - Version: `origin/main` at `0509b53`

**Additional context**
Confirmed P3 impact: this requires an invalid/non-positive clear ID or a narrow in-flight interaction, so it is not a common debugger failure. However, it silently loses client-side breakpoint state and can mutate the adapter's replace-all breakpoint set, making the resulting behavior surprising and difficult to diagnose. The minimal fix is an early `id <= 0` rejection in `clearBreakpoint`, before inspecting or mutating `bpsByFile`.

PR #179, opened 2026-08-09
fix(dapcli): fail requests when the DAP transport dies

### LINKED ISSUE #172: [dapcli] transport death makes blocking requests return success (nil error), and post-death requests stall for the full 15s timeout
**Describe the bug**

`cmd/dapcli`'s request/response plumbing has no connection-death state. Two distinct defects follow from that, both confirmed with a `net.Pipe` repro under `-race`.

**1. In-flight blocking requests report false success after the socket dies.**

`failPending` ([cmd/dapcli/main.go](../blob/main/cmd/dapcli/main.go), `func (h *dapCLI) failPending()`) unblocks waiters by **closing** each pending channel:

```go
func (h *dapCLI) failPending() {
	h.mu.Lock()
	for seq, ch := range h.pending {
		close(ch)
		delete(h.pending, seq)
	}
	h.mu.Unlock()
}
```

But `request` receives **without comma-ok**:

```go
select {
case msg := <-ch:
	...
	if rm, ok := msg.(godap.ResponseMessage); ok { ... }
	return msg, nil          // <-- msg is nil, err is nil
```

A receive from a closed channel yields the zero value (`nil` `godap.Message`). The type assertion fails, the `!rs.Success` branch is skipped, and `request` returns `(nil, nil)` — a **successful** result. Callers then report success on a dead socket:

- `restart` prints `restarted (breakpoints reinstalled)` even though nothing was restarted.
- The startup `initialize` handshake succeeds, so the CLI drops into its REPL against a dead connection instead of exiting with an error.
- `threads` / `bt` / `locals` fall into their `resp, ok := msg.(*godap.XResponse)` guard with `ok == false` and print `(no threads)` / `(no frames)` / `(no locals)` — indistinguishable from a genuinely empty result.
- `setBreakpoints` silently returns without reporting the failure.

**2. A request registered after `readLoop` has already exited waits the full `replyTimeout`.**

`readLoop` returns permanently on the first read error, and nothing records that. A later `request` still increments `seq`, registers itself in `h.pending`, writes to the dead conn (printing only an async `[error] write ...` line), and then blocks the entire `replyTimeout` (15s) before returning a misleading `timeout awaiting X response`. There is no reader left that could ever answer it.

**To Reproduce**

`net.Pipe` repro (dropped into `cmd/dapcli`, run with `go test -tags bingonative -race -run TestRepro ./cmd/dapcli`):

```go
// 1. in-flight: server reads the request, then closes.
server, client := net.Pipe()
h := &dapCLI{conn: client, reader: bufio.NewReader(client),
	pending: map[int]chan godap.Message{}, bpsByFile: map[string][]breakpoint{}}
go h.readLoop()
go func() { _, _ = readDAPMessage(bufio.NewReader(server)); _ = server.Close() }()
m, err := h.request("restart", &godap.RestartRequest{})
// observed: msg=<nil> err=<nil>

// 2. post-death: connection already closed before the request is made.
server, client := net.Pipe()
_ = server.Close()
h := &dapCLI{conn: client, reader: bufio.NewReader(client),
	pending: map[int]chan godap.Message{}, bpsByFile: map[string][]breakpoint{}}
go h.readLoop()
time.Sleep(200 * time.Millisecond) // let readLoop die
start := time.Now()
_, err = h.request("threads", &godap.ThreadsRequest{})
// observed: elapsed=15.002s err=timeout awaiting threads response
```

Observed output:

```
IN-FLIGHT:  msg=<nil> err=<nil>
POST-DEATH: elapsed=15.00227925s msg=<nil> err=timeout awaiting threads response
```

Interactively: start `bingo -dap-addr localhost:4711`, run `dapcli`, kill the server, then type `restart` — it prints `restarted (breakpoints reinstalled)`. Type `threads` — it hangs 15s and then reports a timeout rather than a disconnect.

**Expected behavior**

- A request already in flight when the transport dies unblocks with a **non-nil** connection-closed error that carries the underlying read error, never `(nil, nil)`.
- A request registered after the read loop has died fails immediately (well under 1s) with the same connection-closed error, not after the full 15s reply timeout.
- `restart` / `initialize` / `threads` / `bt` / `locals` surface a clear disconnect error instead of false success or an ambiguous empty result.
- Normal response correlation and the ordinary reply timeout for a live connection are unchanged.
- No send-on-closed-channel or double-close races between concurrent `request`, `onResponse`, `failPending`, and `close` (must be clean under `-race`).

**Locations**

- [cmd/dapcli/main.go](../blob/main/cmd/dapcli/main.go): `readLoop`, `onResponse`, `request`, `failPending`, `close`.

**Additional context**

`pkg/client/ws.go` already has the sound precedent for this: a `done chan struct{}` closed once via `sync.Once` (`signalDone`), with `sendAndWait` selecting on `case <-c.done` to return a real error. `cmd/dapcli` should adopt an equivalent connection-death state (stored `connErr`, registration refused under the same mutex that guards the pending map). If per-request channels remain closed rather than sent-to, the comma-ok form is mandatory.

This is distinct from #128 / #135, which is only about `clear 0` breakpoint-ID validation, and is independent of PR #148 (`pkg/client` timeout reply debt) — no shared code.

**Desktop**
- OS: macOS (darwin) and Linux — platform-independent, pure Go transport logic
- Architecture: arm64 / amd64
- Version: `main` @ 0509b53

xsachax:
## Summary

`cmd/dapcli` had no connection-death state, so a lost DAP transport surfaced as success.

`failPending` unblocked waiters by **closing** each pending channel, but `request` received without comma-ok:

```go
case msg := <-ch:
	if rm, ok := msg.(godap.ResponseMessage); ok { ... }
	return msg, nil   // msg is nil, err is nil
```

A receive from a closed channel yields a nil `godap.Message`; the type assertion fails, the `!rs.Success` branch is skipped, and `request` returns `(nil, nil)`. Callers then reported success on a dead socket:

- `restart` printed `restarted (breakpoints reinstalled)`
- the startup `initialize` succeeded, dropping the operator into a REPL bound to a dead socket
- `threads` / `bt` / `locals` printed `(no threads)` / `(no frames)` / `(no locals)`, indistinguishable from a genuinely empty result

Separately, a request registered **after** `readLoop` had already exited still registered a waiter nobody could answer and blocked the full 15s `replyTimeout` before returning a misleading `timeout awaiting X response`.

## What changed

Explicit connection-death state modelled on the `pkg/client/ws.go` precedent (`done` + `sync.Once` + `signalDone`), scoped to `cmd/dapcli`:

- `markDisconnected(cause)` records the cause and closes a `done` channel exactly once (`sync.Once`), **under the same `mu` that guards the pending-waiter map** — registering a waiter and observing death is one atomic step, so a post-death request fails immediately instead of burning `replyTimeout`.
- **Waiter channels are never closed.** `onResponse` can be mid-send on one, and a closed channel is precisely what produced the nil-message false success. Blocked waiters select on `done` and get a non-nil error carrying the underlying read error.
- That `done` case does a **non-blocking re-check of the waiter channel first**: a server that replies and then closes makes both cases ready at once and `select` picks at random, so a delivered response must still win or correlation is lost.
- `write` now returns its error, so `request` fails fast on a dead send instead of stalling; `fire` keeps its existing async error print.
- `close()` marks death through the same path, so quitting can't strand an in-flight request.
- `newDAPCLI` constructor guarantees `done` is non-nil — a nil channel blocks forever and would silently defeat every death select.

Error text is explicit (`restart failed: connection to DAP server closed: EOF`), never a bare EOF that reads as end-of-stream success. Normal response correlation and the ordinary reply timeout on a live connection are unchanged. No unrelated CLI refactor.

## Tests

Six `net.Pipe` regressions in `cmd/dapcli/main_test.go`:

| Test | Gate |
| --- | --- |
| `TestRequestInFlightFailsWhenConnectionDies` | server reads the request then closes → in-flight request returns a non-nil error, nil message |
| `TestRestartRequestDoesNotSucceedOnDeadConnection` | `restart` at request level cannot succeed; fails in <1s |
| `TestInitializeRequestDoesNotSucceedOnDeadConnection` | handshake cannot enter the REPL on a dead socket |
| `TestRequestAfterReadLoopDeathFailsImmediately` | post-death request completes in <1s, not 15s, and is not a timeout error |
| `TestRequestPrefersDeliveredResponseOverConnectionDeath` | reply-then-close still resolves the response (correlation preserved) |
| `TestConcurrentRequestsSurviveDisconnect` | 16 concurrent requests + `close()` + read-loop death, clean under `-race`; no send-on-closed / double-close |

Verified these actually gate the bug: with the old semantics temporarily restored, 4 of them fail (`TestRequestInFlightFailsWhenConnectionDies` in 0.00s with `err=<nil>`, `TestInitializeRequestDoesNotSucceedOnDeadConnection` and `TestRequestAfterReadLoopDeathFailsImmediately` each after 15.01s). With the fix all 8 tests in the package pass, none exceeding 0.01s.

```
go tool goimports -l cmd/dapcli/   # clean
gofmt -l cmd/dapcli/               # clean
go vet -tags bingonative ./...     # clean
go test -tags bingonative -race -count=1 ./cmd/dapcli   # 8 run / 8 pass / 0 fail
go test -tags bingonative -race -count=25 ./cmd/dapcli  # ok (flake check)
go test -tags bingonative ./cmd/... ./pkg/... ./internal/dap/... ./internal/dapclient/...  # all ok
```

`AGENTS.md` records the new invariant in the same commit, per the repo's doc-sync rule.

## Scope

Independent of PR #148 (`pkg/client` timeout reply debt — no shared code) and distinct from #128/#135, which is only `clear 0` breakpoint-ID validation.

Fixes #172

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1259,6 +1259,24 @@ stays alive for others to join. It buffers `break`s set before launch and flushe
 them on `initialized`. Any mix of `dapcli` and `cli` clients can drive/observe
 one session concurrently.
 
+`dapcli`'s connection death is explicit state, not an implied channel close.
+`markDisconnected` records the cause and closes a `done` channel once
+(`sync.Once`), all under the same `mu` that guards the pending-waiter map, so
+registering a waiter and observing the transport's death is one atomic step —
+a request issued after `readLoop` has already exited fails immediately instead
+of burning the full `replyTimeout`. Waiter channels are **never closed**: a
+receive from a closed channel yields a nil `godap.Message`, whose failed type
+assertion made `request` return `(nil, nil)` and let `restart` print
+"restarted", `initialize` enter the REPL on a dead socket, and
+`threads`/`bt`/`locals` print an empty result — a false success indistinguishable
+from a real one. Blocked waiters instead select on `done` and get a non-nil
+error carrying the underlying read error. That select does a **non-blocking
+re-check of the waiter channel** before reporting death, because a server that
+replies and then closes makes both cases ready at once and `select` picks at
+random; a delivered response must still win or correlation is lost. `close()`
+marks death through the same path, so quitting can't strand an in-flight
+request.
+
 **Option Y (rejected).** The alternative was teaching the hub about DAP directly
 (a second protocol path through `internal/hub`). Rejected: it would fork the most
 fragile fan-out/suspend-gate code for a second protocol. The `hub.WSConn`
--- a/cmd/dapcli/main.go
+++ b/cmd/dapcli/main.go
@@ -54,10 +54,23 @@ type dapCLI struct {
 	conn   net.Conn
 	reader *bufio.Reader
 
-	// mu guards seq and pending (the response-waiter registry).
-	mu      sync.Mutex
-	seq     int
-	pending map[int]chan godap.Message
+	// mu guards seq, pending (the response-waiter registry), and the
+	// connection-death fields. Registering a waiter and observing the
+	// connection's death must be one atomic step, or a request racing the read
+	// loop's exit would register with nobody left to answer it and block for
+	// the whole replyTimeout.
+	mu           sync.Mutex
+	seq          int
+	pending      map[int]chan godap.Message
+	disconnected bool
+	connErr      error
+
+	// done is closed exactly once (closeOnce) when the transport dies, so
+	// waiters already blocked in request unblock. Waiter channels themselves
+	// are never closed: onResponse can be mid-send on one, and a receive from a
+	// closed channel yields a nil message that reads as a successful response.
+	done      chan struct{}
+	closeOnce sync.Once
 
 	// writeMu serialises WriteProtocolMessage across the REPL goroutine, the
 	// initialized-handler goroutine, and any command handler — concurrent writes
@@ -75,6 +88,20 @@ type dapCLI struct {
 	printMu sync.Mutex
 }
 
+// newDAPCLI builds a client over conn. Always construct through this so done is
+// non-nil: a nil channel blocks forever, which would silently defeat every
+// connection-death select.
+func newDAPCLI(conn net.Conn) *dapCLI {
+	return &dapCLI{
+		conn:      conn,
+		reader:    bufio.NewReader(conn),
+		pending:   make(map[int]chan godap.Message),
+		bpsByFile: make(map[string][]breakpoint),
+		curThread: 1,
+		done:      make(chan struct{}),
+	}
+}
+
 func main() {
 	addr := flag.String("addr", "localhost:4711", "DAP server address (host:port)")
 	sessionID := flag.String("session", "", "existing bingo session ID to join (omit to create on launch)")
@@ -86,13 +113,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	h := &dapCLI{
-		conn:      conn,
-		reader:    bufio.NewReader(conn),
-		pending:   make(map[int]chan godap.Message),
-		bpsByFile: make(map[string][]breakpoint),
-		curThread: 1,
-	}
+	h := newDAPCLI(conn)
 	go h.readLoop()
 
 	if *sessionID != "" {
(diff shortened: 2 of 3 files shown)


xsachax:
Superseded by merged #226, which implements transport-death error propagation, immediate post-death request rejection, waiter teardown, and graceful EOF using the current terminalErr/pendingResult lifecycle. The older done/markDisconnected model should not be restacked.

Issue #160, opened 2026-08-09
[hub] Ordinary command overflow silently drops requests and wedges DAP


xsachax:
**Describe the bug**

**Priority: P1.** The hub silently drops every non-resuming command once its bounded command channel fills. At `origin/main` `0509b53`, `internal/hub/hub.go:120` creates `cmdCh` with capacity 32, and `injectCommand` at `internal/hub/hub.go:583-600` uses a non-blocking send with a `default` branch that only logs `command queue full — dropping`. The originating client remains connected and receives no failure.

This breaks the command-delivery contract for all ordinary commands, including `SetBreakpoint`, `ClearBreakpoint`, data requests, `Pause`, `Kill`, and `Restart`. Resume commands are separate: `resumeCh` is intentionally capacity 1 / first-writer-wins and is not part of this bug.

The highest-impact path is DAP. `internal/dap/breakpoints.go:68-81` records every requested breakpoint in `setQ` before enqueuing every `CmdSetBreakpoint`. A single DAP `setBreakpoints` request with about 34+ new lines can deterministically overflow `cmdCh` while one debugger command is slow. Dropped commands never produce `EventBreakpointSet` or `EventError`, so the corresponding slots never resolve and the `setBreakpoints` response never completes. Because confirmations are correlated through id-less FIFOs, later confirmations pop stale entries and permanently shift the adapter's correlation queues.

Control commands are also affected. `Kill` and `Pause` are fire-and-forget in the SDK; a saturated queue can drop either while returning `nil`. DAP can acknowledge terminate/disconnect while the dropped `Kill` leaves a live tracee behind, especially when WebSocket observers keep the shared session alive.

**To Reproduce**

1. Start a hub with one client and gate the fake debugger's first ordinary command so the hub's `Run` goroutine remains inside `executeCommand`.
2. Send one command that enters the debugger, then send at least 33 more non-resuming commands from the same client.
3. The next 32 commands fill `cmdCh`; command 34 and every later command hit the `default` branch in `injectCommand` and are silently discarded.
4. Release the debugger gate. Only the first 33 commands execute; the connection stays open and no error identifies the omitted commands.
5. For the DAP variant, issue one `setBreakpoints` request containing more than 40 new lines. `setQ` contains all slots, but only the admitted prefix can receive confirmations, so no DAP response is sent. A later breakpoint confirmation is consumed by an old slot rather than its own request.
[...]

PR #182, opened 2026-08-09
fix(dap): make setBreakpoints a per-source transaction

### LINKED ISSUE #178: [dap] setBreakpoints is not transactional: replace-all forgets live IDs and pipelined requests resurrect removed breakpoints
## Summary

`internal/dap` treats a `setBreakpoints` request as a fire-and-forget diff instead of a transaction. The adapter's per-source map (`bpByFile`) conflates **what is installed in the debugger** with **what the client last asked for**, and `bpRequest.done()` only tracks the *Set* slots of a request — clears are not first-class pending operations at all.

Two symptoms were independently reproduced. They share one root cause and are filed together because fixing either one alone leaves the other latent (and a partial fix would fight the other's state model).

## Symptom A — a failed clear is invisible, and the live debugger ID is already forgotten

[`internal/dap/breakpoints.go`](https://github.com/bingosuite/bingo/blob/main/internal/dap/breakpoints.go) deletes the line from `bpByFile` **before** the `ClearBreakpoint` is confirmed:

```go
for line, state := range current {
    if desired[line] { continue }
    delete(current, line)                 // mapping dropped up-front
    if state.debuggerID == 0 { continue }
    h.clearQ = append(h.clearQ, state.debuggerID)
    ...
}
...
allResolved := reqObj.done()              // only inspects Set slots
```

Repro (adapter suspended at a breakpoint, `/x/main.go:10` installed as debugger id 101):

1. Client sends `setBreakpoints` for `/x/main.go` with an **empty** `breakpoints` array.
2. `reqObj.slots` is empty → `done()` is trivially true → a **success** `setBreakpointsResponse` is sent immediately, before the `ClearBreakpoint` has been answered.
3. The hub answers with `EventError{Command: clearBreakpoint}` — e.g. `breakpoint clear: restore bytes at 0x…: …` (a memory write, which can fail while the tracee is running) or `breakpoint 101 not found` for a stale/unknown id (see `breakpointTable.clear` in [`internal/debugger/breakpoint.go`](https://github.com/bingosuite/bingo/blob/main/internal/debugger/breakpoint.go)).
4. `onError` routes `CmdClearBreakpoint` to `onBreakpointCleared`, which only pops `clearQ`.

Result: the trap is **still armed** in the debugger, the adapter has already forgotten debugger id 101, the user was told the request succeeded, and no later `setBreakpoints` can ever remove it (the id is gone). Note `breakpointTable.clear` keeps the entry when `WriteMemory` fails, so the breakpoint genuinely survives.

## Symptom B — a pipelined request resurrects a removed breakpoint (and can duplicate a Set)

DAP explicitly permits request pipelining, and VS Code overlaps rapid breakpoint toggles / save-triggered updates. Same-source pipelining breaks:

1. Request **A**: `setBreakpoints /x/main.go [10]` → `current[10] = breakpointState{}` (pending sentinel), `CmdSetBreakpoint` enqueued, no response yet.
2. Request **B**: `setBreakpoints /x/main.go []` → the clear loop sees `debuggerID == 0`, so it `delete`s line 10 and `continue`s **without sending any Clear**; B has no slots → immediate success with an empty body.
3. A's `EventBreakpointSet{ID: 101}` finally arrives. `onBreakpointSet` pops `setQ[0]` and blindly writes `lines[10] = {dapID: 101, debuggerID: 101}` — **resurrecting a line the latest request removed**, and committing it as current desired state.

Result: the breakpoint is armed in the debugger and believed live by the adapter, even though the newest request asked for none. No `ClearBreakpoint` is ever sent.

A second variant of the same root: if a request wants a line whose Set is still pending, the `state.debuggerID != 0` guard fails and a **duplicate** `CmdSetBreakpoint` is issued for the same file:line. `breakpointTable.set` rejects it with `errBreakpointExists`, and the resulting `EventError` runs `failBreakpointSet`, which does `delete(lines, slot.line)` — **deleting the entry that holds the live debugger id from the first, successful set**. Same end state as Symptom A: an armed, unremovable breakpoint.

This is **not** the documented multi-driver FIFO caveat in AGENTS.md — a single DAP client, sole driver of the session, reproduces both.

## Root cause

One transactional defect:

- `bpByFile` stores a single fused record, so "installed in the debugger" and "the client's latest intent" cannot disagree — but during pipelining they legitimately do.
- Clears are not owned by the request that caused them: they have no slot, do not gate the response, and their failure has nowhere to be reported.
- The mapping (`debuggerID`) is deleted on *intent* rather than on *confirmation*, so a failure has nothing left to restore.
- A superseded pending Set has no way to be marked "cancel on confirm", so its confirmation is committed instead of being immediately undone.

## Expected behaviour

`setBreakpoints` should be a per-source, latest-wins transaction over installed facts:

- installed state is only mutated by a confirmation; a clear failure retains the mapping and fails (or otherwise accurately completes) the originating request;
- a response is never sent before every operation that request owns — clears included;
- a pending Set that a later request supersedes is cancelled on confirm and cleared immediately, never committed as desired state;
- an overlapping request wanting an already-pending line attaches to the in-flight operation instead of issuing a duplicate Set;
- exactly one response per request; sources stay independent.

Out of scope: no wire-protocol change, no rejection of overlapping requests, and the id-less confirmation FIFO limitation is preserved.

## Related

- #127 / #138 — restart id reconciliation only; prerequisite, not a duplicate (this builds on its dapID/debuggerID split).
- #160 / #162 — hub command admission; a dependency of the enqueue path, not a duplicate.

xsachax:
Stacked on #138 (base `xsachax-fix-dap-restart-breakpoints`) — its dapID/debuggerID split is a prerequisite.

> **Depends on #162** (`fix(hub): prevent ordinary command loss`, fixing #160). See *Dependency and merge order* below — this PR is **not** standalone-merge-safe to `main` before #162.

## Summary

`setBreakpoints` was a fire-and-forget diff over a single fused per-line record, so "installed in the debugger" and "what the client last asked for" could never disagree — but under pipelining they legitimately do. Two independently reproduced symptoms shared that one root:

- **Replace-all forgot live IDs.** The removed line was deleted from `bpByFile` before its `ClearBreakpoint` was confirmed, and a request's completion check only looked at its Set slots. A remove-all responded success immediately; a rejected clear only popped `clearQ`, leaving the trap armed (`breakpointTable.clear` keeps the entry when the memory write fails), its debugger id forgotten, and the user told it worked.
- **Pipelined requests resurrected removed breakpoints.** `A:[10]` then `B:[]` dropped line 10 without sending a Clear (its id wasn't known yet) and answered B success; A's late confirmation then wrote the line back as live. An overlapping request wanting an already-pending line also issued a duplicate `CmdSetBreakpoint`, whose `errBreakpointExists` failure deleted the entry holding the first, live id.

## What changed

`bpLine` now keeps three things apart per file:line — `installedID` (a fact: written only by a confirmed set, dropped only by a confirmed clear), `desired` (latest-wins intent), and `pending` (the single **live** `bpOp`). `advanceLineLocked` is a convergence loop: issue the one command the gap calls for, or settle everyone parked on the line when the two agree. Every confirmation re-enters it.

That yields the invariants:

- removals are owned by the request that caused them, so a response never precedes its own clears;
- a rejected clear **retains** the mapping and fails the originating request with a DAP error response — and, because it ends the line's only in-flight operation, it also answers anything else parked there rather than stranding it;
- a superseded pending set is cleared the instant its id arrives, never committed as desired state;
- an overlapping request attaches to the in-flight operation instead of duplicating it;
- exactly one response per request (`bpRequest.ready()` claims the right to answer), answered in `reqSeq` order;
- sources stay independent.

Commands are staged in an `outbox` drained in reservation order (`flushCommands`/`flushMu`), because both the DAP read loop and the hub write pump now produce them and the id-less FIFO correlation only works if wire order matches slot-reservation order.

No wire-protocol change, no rejection of overlapping requests, and the documented id-less FIFO limitation is unchanged.

### Restart: re-identification and abandoned operations (commits 2–3)

Restart reconciliation from #138 is preserved, with two corrections the transaction made reachable:

**Re-identify, don't skip.** A line with an operation in flight is re-identified too. Keeping the pre-restart id was a real defect: that id is dead after the relaunch, the already-marshalled clear is rejected by the new engine, and retain-on-failure would then reissue the same dead id forever — leaving the breakpoint the restart actually installed permanently unremovable. Correlation rides the queued `*bpOp`, not `installedID`, so adopting the fresh id cannot desynchronise `setQ`/`clearQ`, and the mapping self-heals before the stale clear fails. A line with `installedID == 0` is still skipped — it owns no identity yet, so a payload entry for it describes another driver's breakpoint.

**Abandon the operation on a discard.** A discarded line previously kept its in-flight clear as `pending` while zeroing everything else — a zombie operation that latched the line. A later request wanting the breakpoint back could issue nothing; another empty replace-all took on a removal obligation for a line already removed; and the eventual stale rejection then discharged both against the wrong state — spuriously failing a removal that had happened, and answering the waiting request unverified without ever attempting its Set, leaving the line `desired` but unarmed with nothing left to drive it. Such an operation is now **abandoned**: it stops being the line's `pending` op (so the line converges again immediately) but keeps its FIFO slot, because the debugger will still answer it and that answer must pop the head it reserved. `liveLineLocked` matches a popped op against the line's current `pending`, so an abandoned or superseded answer does exactly one thing — pop. It never settles waiters, fails owners, or disarms an identity that now belongs to a re-identified line.

`failClearLocked`'s refusal to re-enter the convergence loop is narrowed to the case that justifies it — a still-armed line, where retrying would reissue a failing clear forever. With nothing armed there is nothing to retry and nothing to report, so the removal completes successfully and the line converges on the newest intent.

## Dependency and merge order

This transaction assumes every ordinary command it enqueues eventually produces a confirmation or an error. There are deliberately **no adapter-side timeouts**: one would mask command loss and could fire against a merely slow debugger, so the fix does not invent one.

Under #160 the hub silently drops ordinary commands once `cmdCh` fills. With this change, such a drop no longer merely strands a request — it latches that line's `pending` forever, so no later operation can be issued for it either. #162 is therefore a genuine dependency, not a nice-to-have.

**Merge order: #162 → (#138) → this PR.** #138 and #162 are independent of each other (different packages) and may land in either order, but #182 must not reach `main` before #162. This branch deliberately does **not** duplicate or vendor any part of the hub fix. The adapter's half of the delivery contract is covered here — `TestLargeSetBreakpointsDeliversEveryCommandInOrder` proves the handler blocks rather than drops on a `3 × cmdBufferSize` burst and preserves reservation order; the hub's half is `injectCommand`'s admission in #162. A cross-cutting regression that only passes with #162 present would fail CI on this branch, so it belongs with the hub fix.

## Tests

New `internal/dap/breakpoints_test.go` over the existing loopback-TCP handler harness — 21 specs covering: failed clear after remove-all retains the id and fails the request; partial clear failure; a failed clear still answers the superseded set; a failed clear with nothing armed completes rather than fails and resumes converging; A-set-then-B-empty issues exactly one Clear and ends empty; duplicate pending set emits one Set; set→remove→re-add leaves exactly one live breakpoint; removal completes before a re-add is installed; cross-source independence; restart re-identifying a line under a pending clear (self-healing id) and ignoring a line with no installed identity; a discarded line accepting a later set and ignoring the stale rejection; a discarded line not failing an already-satisfied removal; a late stale clear *success* not disturbing the re-identified line while keeping the FIFO aligned; no-drop/in-order command delivery; exactly-once responses. `TestSetBreakpointsDiffAndFIFO` and the restart specs were updated for the new (correct) "wait for the clear" semantics.

Every new spec was mutation-checked against the behaviour it pins. Two are hardened specifically so the mechanism cannot be removed silently: the stale-rejection spec asserts the abandoned operation's answer is a pure FIFO pop (SetBreakpoint count stays at two, the set queue holds exactly the one live re-add op), which fails if `liveLineLocked` loses its pending-op guard; and the exactly-once spec drives a request that is genuinely re-offered — an already-converged line settles it inside `onSetBreakpoints` and the trailing readiness check then tests it again — which produces a duplicate response if `ready()` stops claiming.

```
gofmt/goimports: clean
golangci-lint run --build-tags bingonative ./internal/dap/...   # 0 issues
go vet -tags bingonative ./...
go vet -tags "e2e bingonative" ./test/integration/
go test -tags bingonative -race -count=1 ./internal/dap/... ./internal/hub/...
go test -tags bingonative -race -count=5 ./internal/dap/
```

AGENTS.md records the transaction invariant, live-vs-abandoned operation semantics, the restart rules, and the ordinary-command delivery dependency.

Fixes #178

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1134,19 +1134,59 @@ documented caveat. Fixing it properly needs correlation ids in the bingo protoco
 need no correlation, so multi-driver continue/step is fine; only the id-less
 confirmation requests are affected.
 
-**setBreakpoints is replace-all** (`breakpoints.go`): diff the requested lines
-for a source against `bpByFile` — clear removed, set new, keep unchanged — and
-respond once every slot in the request resolves, in request order. Clearing the
-breakpoint the process is currently parked on re-arms it through the engine's
-step-off path (see the clearbp spec), so the e2e continue-to-exit uses a
+**setBreakpoints is a replace-all TRANSACTION** (`breakpoints.go`): a request
+records the source's latest-wins intent and is answered only once every operation
+it owns has been confirmed. `bpByFile` is `file → line → *bpLine`, and `bpLine`
+deliberately keeps three things apart, because pipelined requests make them
+legitimately disagree:
+
+- `installedID` — an installed FACT. Written **only** by a confirmed
+  `EventBreakpointSet` and dropped **only** by a confirmed
+  `EventBreakpointCleared`. Never on intent.
+- `desired` — the newest request's intent for that line (latest wins).
+- `pending` — the single in-flight `bpOp` for that line. One op per line at a
+  time is what keeps the id-less `setQ`/`clearQ` FIFOs unambiguous.
+
+`advanceLineLocked` is the convergence loop: it issues the one command the
+installed/desired gap calls for, or settles everyone parked on the line when the
+two already agree. Every confirmation re-enters it, which yields the invariants:
+
+- A **removal is owned by the request that caused it** (`openClears` +
+  `clearOwners`), so a response never precedes its own clears. A remove-all no
+  longer reports success while the `ClearBreakpoint` is still in flight.
+- A **rejected clear retains the mapping** and fails the originating request
+  (`clearFailure` → an error `setBreakpoints` response). Forgetting the id would
+  leave a breakpoint the client can never remove — `breakpointTable.clear` keeps
+  the entry when the memory write fails, so the trap really is still armed. The
+  failure path deliberately does not re-enter the loop; a new request retries.
+- A **pending Set superseded before it confirms** is cancelled on confirm: the id
+  is recorded as a fact (the trap IS armed) but `desired` stays false, so the
+  next advance clears it immediately. It is never committed as desired state.
+- An **overlapping request wanting an already-pending line attaches** to that op
+  (another `setWaiter`) instead of issuing a duplicate `CmdSetBreakpoint`, which
+  the engine would reject with `errBreakpointExists` — and whose failure handling
+  used to delete the entry holding the live id.
+- **Exactly one response per request** (`bpRequest.ready()` claims the right to
+  answer; slots and obligations settle on different goroutines), and requests are
+  answered in `reqSeq` order so a later request's view lands last.
+- Sources are independent; the id-less FIFO limitation is unchanged.
+
+Breakpoint commands are staged in `outbox` and drained by `flushCommands`
+(serialised by `flushMu`) rather than enqueued directly: both the DAP read loop
+and the hub write pump produce them, and the FIFOs only correlate if the wire
+order matches the order slots were reserved under `mu`.
+
+Clearing the breakpoint the process is currently parked on re-arms it through the
+engine's step-off path (see the clearbp spec), so the e2e continue-to-exit uses a
 no-breakpoint target, not a clear-then-continue.
 
-`bpByFile` keeps the stable DAP id separate from the debugger's internal id.
-After `EventRestarted`, reconcile its exact source-path/line keys from
-`RestartedPayload` before replying: retained entries adopt the fresh debugger
-id, while discarded entries are removed and emit a `breakpoint` changed event
-with their prior DAP id and `verified:false`. Do not reset `setQ`/`clearQ`;
-those FIFOs still own any in-flight confirmations.
+`bpLine` keeps the stable DAP id (`dapID`) separate from the debugger's internal
+id (`installedID`). After `EventRestarted`, reconcile its exact source-path/line
+keys from `RestartedPayload` before replying: retained entries adopt the fresh
+debugger id, while discarded entries are removed and emit a `breakpoint` changed
+event with their prior DAP id and `verified:false`. A line with an op in flight
+is skipped — that op still owns the line and its confirmation is still queued.
+Do not reset `setQ`/`clearQ`; those FIFOs still own any in-flight confirmations.
 
 ### Server wiring + multi-client discovery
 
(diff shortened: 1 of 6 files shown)


xsachax:
Superseded by merged PR #203, which landed this transactional setBreakpoints layer together with stable logical breakpoint IDs and full validation.

Issue #174, opened 2026-08-09
[hub] Rejected Restart or Kill abandons the suspended wait loop


xsachax:
**Describe the bug**

**Severity: P2.** At `origin/main` `0509b53bfc44da889e152ee20f56ce3663882c65`, the suspended wait loop in `Hub.handleEvent` decides whether to abandon the suspend based **only on the command kind**, never on whether the command actually did anything:

```go
// internal/hub/hub.go — handleEvent's suspend-wait loop
case cc := <-h.cmdCh:
    // Non-resuming command (SetBreakpoint, Locals, …) while suspended.
    // Execute immediately — process is paused — and keep waiting.
    // Restart and Kill are the exceptions: Restart tears down and
    // replaces the suspended process, and Kill terminates it, so in
    // both cases the process we were waiting to resume no longer
    // exists — return and let Run's outer loop pick up the new or
    // closed debugger's events channel.
    h.executeCommand(cc.cmd)
    if cc.cmd.Kind == protocol.CmdRestart || cc.cmd.Kind == protocol.CmdKill {
        return
    }
```

The comment's premise ("the process we were waiting to resume no longer exists") only holds when the command *succeeded*. Several paths reach `return` while the suspended process is still alive and the hub is still `StateSuspended`.

`Restart` is rejected **before** it touches `h.dbg` or the session state in three places in `handleRestart`:

```go
if h.sessionID == "" || h.newDebugger == nil {
    h.broadcastError(cmd.Kind, fmt.Errorf("restart requires a managed session"))
    return                                    // raw hub (hub.New) — no factory
}
if h.lastLaunch == nil {
    h.broadcastError(cmd.Kind, fmt.Errorf("no launched process to restart — use Launch first"))
    return                                    // Attach-based session, or no prior Launch
}
var override protocol.RestartPayload
if len(cmd.Payload) > 0 {
    if err := protocol.DecodeCommandPayload(cmd, &override); err != nil {
        h.broadcastError(cmd.Kind, err)
        return                                // malformed RestartPayload
    }
}
```

`Kill` fails the same way through `executeCommand` → `dispatch` → `dbg.Kill()`:

```go
result, err := dispatch(h.dbg, cmd)
if err != nil {
    h.log.Warn("command failed", "kind", cmd.Kind, "err", err)
    ...
    h.broadcastError(cmd.Kind, err)
    return                                    // state untouched: still StateSuspended
}
```

[...]

Issue #158, opened 2026-08-09
[hub] Suspend timeout bypasses resume state transitions and strands failed retries


xsachax:
**Describe the bug**

**Severity: P2.** At `origin/main` `0509b53bfc44da889e152ee20f56ce3663882c65`, the suspended wait loop's 30-minute safety timeout calls `h.dbg.Continue()` directly and then unconditionally returns:

```go
case <-timeout.C:
    h.log.Warn("30-minute suspend timeout — auto-continuing")
    if h.dbg != nil {
        if err := h.dbg.Continue(); err != nil {
            h.log.Warn("auto-continue failed", "err", err)
        }
    }
    return
```

This bypasses `executeCommand`, which is the path that both dispatches a resume and transitions the hub's managed session state to `running` only after the debugger accepts it.

Exact success trace:

1. A suspending event is broadcast and `handleEvent` transitions the hub to `StateSuspended`.
2. After 30 minutes, the timeout calls `Debugger.Continue()` directly.
3. The debugger succeeds and emits exactly one `EventContinued`, but the timeout branch returns without `transitionState(StateRunning)` and therefore without a managed `EventSessionState{State: running}`.
4. `Run` later forwards the non-suspending `EventContinued` while the hub's state remains `suspended`, so the tracee is running but `Hub.State()`, `/api/sessions`, and future welcome events report `suspended`.
5. A DAP client joining the session consumes that stale welcome as a real suspend and emits a bogus `stopped reason=pause` even though the target is running. The next real stop can further desynchronize state-driven clients because the hub is already recorded as suspended and emits no suspended state transition.

Exact failure trace:

1. The timeout calls `Debugger.Continue()` directly and the debugger rejects it while remaining suspended (for example, a transient breakpoint-reinstall failure).
2. The branch logs the error only; it emits no protocol `EventError`, then unconditionally returns from the suspended wait loop.
3. `Run`'s outer select never drains `resumeCh`, so every later client `Continue`/`Step*` retry is buffered or dropped and never reaches the debugger. Only `Kill` or `Restart` can recover the session.

This duplicates the invariant gap fixed for client-triggered rejected resumes in #95, but through the still-separate timeout branch. #104 concerns the distinct stale-resume drain ordering race and is not a duplicate.

**To Reproduce**

Runtime reproduction:
1. Start a managed session and stop on a breakpoint, pause, panic, or step completion.
2. Leave it suspended for 30 minutes.
[...]

PR #184, opened 2026-08-09
fix(debugger): suspend the hub on asynchronous engine halts

### LINKED ISSUE #183: [debugger] Asynchronous halts in handleStop are non-suspending, permanently stranding the session
**Describe the bug**

**Severity: P2.** At `origin/main` `0509b53bfc44da889e152ee20f56ce3663882c65`, six error paths in `engine.handleStop` halt the tracee and set `stateSuspended`, but report the halt with only a **non-suspending** `EventError`. Because these failures happen *asynchronously* — long after the resume command was answered `nil` and `EventContinued` was broadcast — the hub has already transitioned to `running` and left its suspend wait loop. `resumeCh` is drained **only** inside that wait loop (`Run`'s outer `select` never reads it), so every subsequent `Continue`/`Step*` sits unread in a capacity-1 channel. The tracee is halted forever, the hub reports `running`, and the session can never be resumed again.

This is a genuine engine-side gap, not a duplicate:

- **#174 / #180** (`[hub] Rejected Restart or Kill abandons the suspended wait loop`) and **#175** (`[dap] Rejected Continue, Step, or Restart…`) both concern **synchronous** rejection — the resume returns a non-nil error, so the client/adapter/hub can observe the failure on the command's own return path. Here the resume **succeeds** (`Continue` returns `nil`), and the failure surfaces later on the engine loop with no command in flight. Nothing downstream can attribute it.
- **#158** concerns the 30-minute suspend timeout, a different code path.

The engine already relies on this invariant elsewhere and states it explicitly in `emitStoppedAtCurrentPC`:

```go
// Always emits even on register-read failure: the hub needs
// a suspending event or it loses track of state and drops resume commands.
```

`handleStop`'s error paths violate exactly that rule.

**Affected sites** (all in `internal/debugger/engine.go`, `handleStop`)

| Site | Location | Trigger | State on exit |
| --- | --- | --- | --- |
| A | ~590-596 `StopBreakpoint` | `populateBreakpointStop` fails | `stateSuspended` |
| B | ~640-651 `StopSingleStep` | `populateStopPC` fails | `stateSuspended` |
| C | ~657-665 `StopSingleStep` | `bps.reinstall` fails | `stateSuspended` |
| D | ~713-718 `bpResumeStepOut` | temporary return BP `bps.set` fails | **`stateRunning`** (wrong) |
| E | ~731-738 `StopSignal` | in-flight step-over `bps.reinstall` fails | `stateSuspended` |
| F | ~747-752 `StopSignal` (manual Pause) | `populateStopPC` fails | `stateSuspended` |

Site **D** is strictly worse than the rest: it never calls `setState(stateSuspended)` at all, so the engine believes the process is running while the tracee is halted with the trap reinstalled and threads released. Every later `Continue`/`Locals`/`StackFrames` is then rejected with `ErrNotSuspended`, and `Pause` is the only command that would be accepted — but no interrupt can ever arrive, because nothing is executing.

The source-step fallback (`bpResumeSourceStep`) is already correct: it emits a suspending `EventStepped`.

**Exact event/state trace (site C, the canonical case)**

1. Tracee suspended at a user breakpoint; hub is `suspended` and parked in `handleEvent`'s wait loop.
2. Client sends `CmdContinue`. The hub drains `resumeCh` and calls `Debugger.Continue()`.
3. `engine.Continue` → `resumeFromBreakpoint(bpResumeContinue, 0)`: restores the original bytes, arms the single-step, `setState(stateRunning)`, starts `waitLoop`, **returns `nil`**.
4. `engine.Continue` emits `EventContinued`. `executeCommand` sees success → `transitionState(StateRunning)` → `h.State() != StateSuspended` → the wait loop **returns**.
5. The single-step completes. `handleStop` → `StopSingleStep` → `bps.reinstall` fails (transient backend write error).
6. Engine correctly refuses to resume a trap-less process: `endThreadStep()`, `setState(stateSuspended)`, `emitError(protocol.CmdNone, …)`, `return`.
7. `EventError` is **not** in `suspendingEvents`, so the hub broadcasts it and stays `running`. No wait loop is entered.
8. Client sends `CmdContinue` again. `injectCommand` routes it to `resumeCh` (capacity 1). Nothing drains it. **The engine never sees it.**
9. Repeat forever. `Hub.State()` and `/api/sessions` report `running`; a DAP client shows a running program that can never be stepped, inspected, or resumed. The 30-minute safety timeout does not help — it lives in the wait loop that was never re-entered.

**To Reproduce**

Deterministic, no OS backend involved — a fault-injected fake `Backend` driving the **real** engine and the **real** hub:

1. `fakeBackend` gains fault injection (`failWriteAt`, `failReadAt`, `failRegisters`, `failThreads`), guarded by a mutex and armed before the triggering stop is pushed.
2. Engine-only: set a breakpoint, push `StopBreakpoint`, `Continue()` (consuming `EventContinued`), arm `failWriteAt(bpAddr)`, push `StopSingleStep`. Observed: exactly one `EventError`, then nothing. A subsequent `Continue()` still succeeds at the engine level, proving the engine itself recovered — the loss is purely in what the hub was told.
3. Real hub: `hub.New(realEngine, nil)` + a fake `WSConn`, same sequence driven over the wire. After the injected failure, `Hub.State()` remains `StateRunning` and a second `CmdContinue` never reaches `Backend.ContinueProcess`.

All six sites reproduce identically: the detailed `EventError` arrives with the correct message, and no suspending event ever follows.

**Expected behavior**

An asynchronous halt must be reported like any other suspend, so the hub re-enters its wait loop and later resumes are drained:

1. Preserve the detailed `EventError` first (it carries the actual cause and must not be diluted).
2. Ensure `stateSuspended` — site D must add the missing `setState`.
3. Emit a **suspending** `EventPaused`.

Ordering is `Continued → Error → Paused`. Emitting `Paused` last is load-bearing: the hub drains `resumeCh` *before* broadcasting a suspending event, so any legitimate retry the client sends in response to `Paused` necessarily lands after that drain and cannot be eaten as a stale resume.

`EventPaused` is the correct kind — it already means "halted asynchronously at an arbitrary instruction, not a source-line boundary", it is already in `suspendingEvents`, and DAP already maps it to `stopped` reason=`pause`. The alternatives are all wrong here: `EventStepped` is treated as the entry stop by the DAP restart path and can trigger an auto-continue; `EventBreakpointHit` would claim a trap that may not be installed (site C/E is precisely a failed reinstall); an unsolicited `EventBreakpointCleared` would corrupt the adapter's id-less FIFO correlation.

The halt event must be built **cheaply** — a synthetic goroutine rather than a full `goroutineSnapshot`, since the backend is by definition already failing and a snapshot issues dozens of further reads through it.

Notes:

- The failed-reinstall breakpoint must stay absent from the table. Re-adding an entry whose trap is not guaranteed to be installed would report an armed breakpoint that cannot fire.
- No wire-protocol change: no new event kind, no version bump.
- No hub or DAP production change is required — `EventPaused` is already suspending and already translated.

**Desktop**

- OS: macOS (darwin/arm64) and Linux (linux/amd64) — platform-independent
- Architecture: arm64, x86_64
- Version: `origin/main` `0509b53bfc44da889e152ee20f56ce3663882c65`

**Additional context**

**No OS backend impact.** The defect and its fix are entirely in the shared, platform-agnostic `handleStop` state machine. No `backend_darwin_arm64.go`, `backend_linux_amd64.go`, or `trap_*.go` change is involved, so no Darwin verification gate is triggered.

Related but distinct: #174/#180 and #175 (synchronous rejection), #158 (suspend timeout).

Separately observed while confirming site A reachability, **not** part of this report: a `StopBreakpoint` arriving while `steppingOverBP != nil` (a sibling thread trapping during an in-flight step-over) returns through the `StopBreakpoint` branch without reinstalling the step-over breakpoint or calling `endThreadStep()`. That is pre-existing on both the success and failure paths of that branch and needs its own reproduction before any change.

xsachax:
Fixes #183

## The bug

Six error paths in `engine.handleStop` halted the tracee and set `stateSuspended`, but reported the halt with only a **non-suspending** `EventError`.

These failures are *asynchronous*. The resume that led to them — `Continue`/`Step*` off a software breakpoint — already returned `nil` as soon as the step-over single-step was armed, and already emitted `EventContinued`. By the time the failure happens on the engine loop, the hub has transitioned to `running` and **left its suspend wait loop**. Since `resumeCh` is drained *only* inside that loop (`Run`'s outer select never reads it), every later `Continue`/`Step*` sits unread in a capacity-1 channel.

Result: the tracee is halted forever, while `Hub.State()`, `/api/sessions` and DAP all report `running`. The 30-minute safety timeout doesn't help — it lives in the loop that was never re-entered.

Trace (site C, canonical):

```
suspended at breakpoint → hub in wait loop
  CmdContinue → resumeFromBreakpoint arms single-step → returns nil
  EventContinued → hub transitions running → wait loop RETURNS
  single-step completes → bps.reinstall fails
  setState(stateSuspended) + EventError        ← non-suspending
  hub stays running, never waits, never drains resumeCh
  CmdContinue → resumeCh → unread → stranded forever
```

| Site | Trigger | State on exit (before) |
| --- | --- | --- |
| A | `StopBreakpoint` `populateBreakpointStop` fails | suspended |
| B | `StopSingleStep` `populateStopPC` fails | suspended |
| C | `StopSingleStep` `bps.reinstall` fails | suspended |
| D | `bpResumeStepOut` return-BP `set` fails | **running** (wrong) |
| E | `StopSignal` in-flight reinstall fails | suspended |
| F | `StopSignal` manual-Pause `populateStopPC` fails | suspended |

Site D was strictly worse: it never called `setState(stateSuspended)`, so the engine believed the process was running while it sat halted with the step-over trap reinstalled and the world still held — every later command rejected with `ErrNotSuspended` against a process that could not move. (`endThreadStep` ends the stepping bookkeeping and disarms the hardware single-step; it deliberately does *not* resume the other threads — only `ContinueProcess` does, and it is never reached on that path.)

## The fix

`haltOnError(cmd, cause, stop)` reports these halts like any other suspend: the detailed `EventError` first, then a suspending `EventPaused`. The hub re-enters its wait loop and the retry is delivered. Site D also gains the missing `setState(stateSuspended)`.

Order is `Continued → Error → Paused`, and `Paused` must be last: the hub drains `resumeCh` *before* broadcasting a suspending event, so a retry sent in response to `Paused` necessarily lands after that drain and cannot be eaten as a stale resume.

`emitHaltedOnError` issues **no backend calls at all** — no `goroutineSnapshot` (as `emitPaused` does) and not even a stack walk. The backend on this path is by definition already failing, so any read could delay or prevent the one event that restores liveness. Both the location and the synthetic goroutine go through `locForPC`, so a PC of 0 short-circuits instead of scanning every CU and the two always agree.

### Delivering the suspend is guaranteed by a reserved slot

`emit` is non-blocking and drops on a full buffer, so a saturated buffer could lose the `Paused` and recreate #183 — deterministically at zero free slots, where the cause was skipped *and* the suspend dropped, leaving the hub with nothing at all.

The events channel is now `eventBufSize + eventBufReserve` deep. Ordinary `emit` refuses to use the last slot; only `emitReserved`, used solely by the halt path, may. `haltOnError` emits the cause as an ordinary event — logging it instead when no ordinary slot is free, since losing detail is far cheaper than losing the suspend — and the `Paused` through `emitReserved`.

**One reserved slot suffices**: a second halt can never be queued behind the first. Emitting a halt leaves the engine `stateSuspended` with the tracee stopped; a stopped tracee cannot produce the next stop; and the only route back to running is a resume, which the hub sends exclusively from inside the suspend wait loop it enters *after receiving* that event. The same argument covers the ordinary suspending events — none can still be buffered when a halt is handled.

Eviction was considered and rejected: removing a queued event requires receiving it, and putting a suspending one back would reorder the entire buffer behind it. The length check needs no lock — the loop thread is the sole writer, so `len` can only fall as the hub drains, and observing room guarantees the send succeeds. The engine loop stays non-blocking throughout.

### `curTID` must not inherit a guessed thread

`populateStopPC` committed `Threads()[0]` into `stop.TID` *before* confirming it with a register read, so a TID-less stop whose `GetRegisters` then failed handed that guess back — and the halt event adopted it into `curTID`, the thread every later step primitive targets. On darwin `task_threads` returns creation order, so `threads[0]` is frequently an idle runtime M. The guess is now kept local until `GetRegisters` succeeds; a real nonzero backend TID is preserved on either path, and an unresolved stop leaves `curTID` alone.

### Why `EventPaused`, and not something else

- **`EventStepped`** — DAP's restart path treats the first Stepped as the relaunched process's entry stop and, without `stopOnEntry`, auto-continues past it. A halt must never auto-resume. Pinned by a test.
- **`EventBreakpointHit`** — would claim a trap that may not be installed (sites C/E are precisely a failed reinstall).
- **unsolicited `EventBreakpointCleared`** — would corrupt DAP's id-less FIFO correlation.
- **making all `EventError`s suspending** — wrong; most errors do not halt anything.

No wire-protocol bump, no new event kind. A failed-reinstall breakpoint stays **out** of the table — re-adding it would advertise an armed breakpoint that can never fire — and the error text says so.

## Scope

- **No OS backend changes.** `backend_darwin_arm64.go`, `backend_linux_amd64.go`, `trap_*.go` untouched, so the Darwin verification gate is not triggered.
- **No hub or DAP production changes.** Those diffs are comments only, including correcting one in `hub.go` that described the reinstall failure as a *synchronous* rejected resume — it is not, and that is exactly why the wait loop could not catch it.
- Synchronous rejection behaviour is unchanged (#174/#175/#180 cover that; #158 covers the timeout).

## Tests

`fakeBackend` gains mutex-guarded fault injection (`failWriteAt`/`failReadAt`/`failRegisters`/`failThreads`); faults are armed before the triggering stop is pushed, so the arming is ordered ahead of the engine's read.

- `internal/debugger/engine_halt_test.go` — one spec per site A–F, each asserting `Error` then `Paused` **and** that a subsequent `Continue` reaches the backend; a spec pinning that an unresolved stop TID does not replace `curTID`; and a saturation spec that fills every ordinary slot, drives a **real** injected step-over failure, and proves both that the session re-enters the suspended gate (a retried `Continue` succeeds and reaches `ContinueProcess`) and that the `Paused` survived in the buffer. Ablating `eventBufReserve` to 0 fails it.
- A **real `Hub` + real engine** over the fault-injected backend proves the end-to-end property: after the injected failure the hub reports `suspended`, and a second `CmdContinue` actually reaches `Backend.ContinueProcess`. It lives in `internal/debugger` because the hub's own `fakeDebugger` cannot exercise asynchronous stop handling.
- `internal/dap/handler_test.go` — the halt surfaces as `stopped` reason=`pause` and leaves the session inspectable, and a halt during `restarting` is **not** treated as an entry stop (verified to fail if `Stepped` were used).

Every new test was verified non-vacuous by reverting the corresponding production change. Verified with `gofmt`/`goimports`, `go vet -tags bingonative`, and `go test -tags bingonative -race` on `internal/debugger`, `internal/hub`, `internal/dap` and `pkg/protocol` over repeated passes, plus randomized-order (`-ginkgo.randomize-all`) runs of the debugger suite.

## Reviews and known residuals

Two independent adversarial reviews were run, plus follow-up rounds. Their findings on `collectFrames` cost, the buffer-drop hole (both the one-slot and the zero-slot case), the `curTID` guess, an inaccurate StepOut comment, and the `locForPC` inconsistency are all incorporated above.

Deliberately **not** fixed here, and explicitly **not** claimed closed:

1. **A halt promises control, not a pristine tracee.** At site A there is no located stop, so the PC was never rewound off the trap: on amd64 RIP sits one byte past the `INT3` and a plain `Continue` would execute from mid-instruction. **`Kill`/`Restart` is the safe recovery there, not `Continue`** — that site's `EventError` now says so, so clients can see it. Elsewhere a partially applied breakpoint write can leave an untracked trap. Reconciling tracee byte state after such a write needs its own design; disabled-breakpoint protocol/state was explicitly out of scope.
2. **Five internal `ContinueProcess` errors are still ignored** in `handleStop` — the spurious-trap advance, `bpResumeContinue`, the successful `StepOut` continuation, the leftover-pause-signal suppression, and the ordinary-signal auto-resume. If one fails, the engine records `stateRunning` for a tracee that never resumed and emits nothing: the same class of liveness loss from the other direction. None has a confirmed backend reachability, so none is fixed and **no issue is filed yet**; this class is *not* globally closed by this PR.
3. A `StopBreakpoint` arriving while `steppingOverBP != nil` returns without reinstalling the step-over breakpoint or calling `endThreadStep()`. **Pre-existing on both the success and failure branches**, not the cause of the liveness loss fixed here, and no deterministic reproduction yet.
4. Two **pre-existing** comments carry a "releases the threads held" inaccuracy: the `endThreadStep` doc comment (`engine.go:141`) and the `StopSingleStep` rearm comment (`engine.go:642`). Left untouched to keep this diff scoped; worth a separate cleanup.

`AGENTS.md` is updated alongside: the rejected-resume and step-over invariants, the reserved-slot guarantee and why one slot suffices, the snapshot-cadence exception, the `handleStop` error-path rule, the safe-recovery guidance, the still-ignored `ContinueProcess` class, and the fault-injection test notes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -342,12 +342,24 @@ that can never fire.
 
 What a halt promises is **control**, not a pristine tracee: the operator regains
 the ability to inspect, retry, or kill. It does not guarantee the tracee's byte
-state is consistent — a write that failed partway can leave an untracked trap,
-and at the `populateBreakpointStop` site the PC was never rewound — so a
-subsequent resume is best-effort and the error text says the breakpoint may no
-longer be armed or tracked. Reconciling tracee state after a partially-applied
-breakpoint write is a separate problem; do not add disabled-breakpoint
+state is consistent, and at the `populateBreakpointStop` site it specifically is
+not — with no located stop the PC was never rewound off the trap, so on amd64
+RIP sits one byte past the `INT3` and a plain `Continue` would execute from
+mid-instruction. **`Kill`/`Restart` is the safe recovery there, not `Continue`**;
+that site's `EventError` says so explicitly. Elsewhere a failed or partially
+applied breakpoint write can leave an untracked trap behind. Reconciling tracee
+state after such a write is a separate problem — do not add disabled-breakpoint
 protocol/state here.
+
+Note also that this rule covers the *reporting* of halts, not every internal
+resume: five `handleStop` paths still ignore `ContinueProcess`'s error (the
+spurious-trap advance, `bpResumeContinue`, the successful `StepOut`
+continuation, the leftover-pause-signal suppression, and the ordinary-signal
+auto-resume). If one of those ever fails, the engine records `stateRunning` for
+a tracee that never resumed and no event is emitted at all — the same class of
+liveness loss from the other direction. None has a confirmed backend
+reachability, so none is fixed or filed yet; **do not treat this class as
+globally closed.**
 
 The sites are `populateBreakpointStop` failure (`StopBreakpoint`),
 `populateStopPC` failure and `bps.reinstall` failure (`StopSingleStep`), the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -187,13 +187,20 @@ When multiple clients race resume commands: **first writer wins**, the rest
 are dropped (`resumeCh` has capacity 1; see [hub.go injectCommand](internal/hub/hub.go)).
 
 Rejected resumes: a resuming command only ends the suspend if the debugger
-actually resumes the process. If the resume is **rejected** — e.g. a transient
-backend error while reinstalling a software breakpoint leaves the engine
-`stateSuspended` — the hub broadcasts the `EventError` but **stays in the wait
-loop** (it checks that the session left `suspended` before returning). Bailing
-out on a failed resume would strand the client: the process is still suspended,
-but a retry resume lands in `resumeCh`, which only the wait loop drains, so the
-session could never be resumed again.
+actually resumes the process. If the resume is **rejected synchronously** —
+e.g. the dispatch returns an error and leaves the engine `stateSuspended` — the
+hub broadcasts the `EventError` but **stays in the wait loop** (it checks that
+the session left `suspended` before returning). Bailing out on a failed resume
+would strand the client: the process is still suspended, but a retry resume
+lands in `resumeCh`, which only the wait loop drains, so the session could never
+be resumed again.
+
+Resumes that are **accepted and only fail later** are a different problem, and
+the wait loop cannot catch them: `Continue`/`Step*` off a software breakpoint
+return `nil` as soon as the step-over single-step is armed, so the hub has
+already transitioned to `running` and left the wait loop by the time the
+failure happens on the engine loop. Those asynchronous halts are reported with a
+suspending `EventPaused` — see [step-over flow](#software-breakpoint-step-over-flow).
 
 ### Session state machine
 
@@ -296,6 +303,59 @@ Internal sentinel BP files: `<stepover-next>`, `<stepout-return>`,
 If `bps.reinstall` ever fails after a single-step, **suspend instead of
 resuming**. Running without the trap is a runaway process; reporting the
 error lets the operator intervene.
+
+**Every asynchronous halt in `handleStop` must be reported with a *suspending*
+event, not a bare `EventError`.** These failures happen after the resume that
+led to them already returned `nil` and emitted `EventContinued`, so the hub has
+transitioned to `running` and left its suspend wait loop. `EventError` is not
+suspending, so on its own it leaves the hub believing the tracee runs while it
+is halted — and since `resumeCh` is drained *only* inside that wait loop, every
+later `Continue`/`Step*` sits unread and the session is stranded for good
+(issue #183). The rule at each such site is:
+
+1. emit the detailed `EventError` (it carries the real cause),
+2. ensure `setState(stateSuspended)`,
+3. emit the suspending `EventPaused`.
+
+`haltOnError(cmd, cause, stop)` does all three in the right order — use it
+rather than open-coding the pair. Order matters: `Continued → Error → Paused`.
+`Paused` must be last because the hub drains `resumeCh` *before* broadcasting a
+suspending event, so a legitimate retry sent in response to `Paused`
+necessarily lands after that drain.
+
+`emitHaltedOnError` emits `EventPaused` with a **synthetic** goroutine and a
+pure-DWARF location, and issues **no backend calls at all** — not
+`emitPaused`/`goroutineSnapshot`, and not even a stack walk. The backend on this
+path is by definition already failing, so any read could delay or prevent the
+one event that restores liveness; clients can ask for frames or a snapshot once
+the session is responsive again. `haltOnError` wraps the pair and, because
+`emit` drops events when the buffer is full, gives the last free slot to the
+`Paused` rather than the `Error` — losing the suspend is what strands the
+session, losing the cause merely loses detail.
+
+`EventPaused` is the right kind: `EventStepped` is treated as the entry stop by
+the DAP restart path and can auto-continue, `EventBreakpointHit` would claim a
+trap that may not be installed, and an unsolicited `EventBreakpointCleared`
+would corrupt DAP's id-less FIFO correlation. A failed-reinstall breakpoint
+stays **out** of the table: re-adding it would advertise an armed breakpoint
+that can never fire.
+
+What a halt promises is **control**, not a pristine tracee: the operator regains
+the ability to inspect, retry, or kill. It does not guarantee the tracee's byte
+state is consistent — a write that failed partway can leave an untracked trap,
+and at the `populateBreakpointStop` site the PC was never rewound — so a
+subsequent resume is best-effort and the error text says the breakpoint may no
+longer be armed or tracked. Reconciling tracee state after a partially-applied
+breakpoint write is a separate problem; do not add disabled-breakpoint
+protocol/state here.
+
+The sites are `populateBreakpointStop` failure (`StopBreakpoint`),
+`populateStopPC` failure and `bps.reinstall` failure (`StopSingleStep`), the
+`bpResumeStepOut` return-breakpoint `set` failure (which is also the one site
+that must add the otherwise-missing `setState(stateSuspended)` — it is reached
+with the engine still `stateRunning`), and the in-flight reinstall and manual
+`populateStopPC` failures (`StopSignal`). The `bpResumeSourceStep` fallback is
+already safe: it emits a suspending `EventStepped`.
 
 ## Architecture-specific traps
 
@@ -624,7 +684,12 @@ auto-emitted on exactly the suspends that can change the concurrency picture —
 single-step/step-over path from extra per-step memory reads. `emitBreakpointHit`
 / `emitPaused` build the snapshot **once**, embed its current goroutine in the
 stop event, then stream the same snapshot — one build, no double scan, no double
-delta pass.
+delta pass. The one `EventPaused` that carries **no** snapshot is the internal
+async halt (`emitHaltedOnError`, see [step-over flow](#software-breakpoint-step-over-flow)):
+it is emitted on a backend error path, where a snapshot would push dozens of
+further reads through the backend that just failed. Observers keep their last
+good model; the halt is still reported, and `CmdGoroutineSnapshot` can refresh
+on demand.
 
 **Not a suspending event.** It follows a suspending event (or answers a query)
 and never gates the hub. DAP `translateEvent` **deliberately ignores** it:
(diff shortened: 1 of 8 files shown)


xsachax:
## Required stack rebase

The user-confirmed published order is **#232 -> #184 -> #202 -> #225**. #232 is frozen at `f37659002448e70c17105ee18d8516eeb27eb46c` on `xsachax-audit-linux-stop-tid-race`.

Please rebase #184's three commits onto that head, force-push `xsachax-fix-async-engine-suspension-halt`, and change this PR's base from `main` to `xsachax-audit-linux-stop-tid-race`. After publishing, verify:

```sh
git merge-base --is-ancestor f37659002448e70c17105ee18d8516eeb27eb46c origin/xsachax-fix-async-engine-suspension-halt
```

#202 will then rebase onto the new #184 head and drop its duplicate atomic commit `04f7f9a`. #225 must wait for that frozen #202 SHA.

#232's final hardening explicitly seeds a launched process in the running-Kill race regression so #204's later attached quiesce path cannot make the inherited test vacuous. Native focused `-race` run `31343912384`, job `93322176089`, is green.

xsachax:
Restacked onto #232 as requested.

- base: `xsachax-audit-linux-stop-tid-race` / `f37659002448e70c17105ee18d8516eeb27eb46c`
- new head: `3e3988d21f6ac92eac95dbe0e1167b3ec53fa9ea`
- `git range-diff` maps all four #184 commits 1:1
- published ancestry: `f376590` is an ancestor of `3e3988d`

Validated before the force-with-lease update with the full Go race suite, `go vet -tags bingonative ./...`, and Linux/amd64 debugger-test cross-compilation.

xsachax:
Superseded by merged PR #202, which landed the asynchronous halt/suspend behavior as part of the validated Linux stop-ownership stack.

PR #189, opened 2026-08-09
fix(vscode): enter filtered tree from the matching end

### LINKED ISSUE #185: [vscode] Reverse tree navigation skips the last visible goroutine when the selection is filtered out
**Describe the bug**
The Bingo Concurrency tree keyboard handler in `editors/vscode/src/webviewApp.ts:376-395` resolves the current position with `visible.findIndex((node) => node.goroutine.id === session.selectedGoroutine)` and then advances with a single modulo expression:

```ts
const direction = event.key === "ArrowDown" || event.key === "ArrowRight" ? 1 : -1;
const next = visible[(current + direction + visible.length) % visible.length];
```

The filter narrows the rendered tree, but the selection is preserved even when the selected goroutine is no longer among the visible nodes. In that state `findIndex` returns `-1`, and the modulo arithmetic is only coincidentally correct in one direction:

- Forward (`ArrowDown`/`ArrowRight`, `direction = 1`): `(-1 + 1 + len) % len === 0`, which selects the first visible node. This is the intended behavior.
- Reverse (`ArrowUp`/`ArrowLeft`, `direction = -1`): `(-1 - 1 + len) % len === (len - 2) % len`, which selects the **second-to-last** visible node and silently skips the last one.

The `-1` sentinel is being treated as a real index one position before the start of the list, so the reverse step lands two positions before the end instead of wrapping to it. The last visible goroutine becomes unreachable by a single reverse keypress, which is exactly the position a user expects when entering a list backwards.

**To Reproduce**
1. Start a bingo debug session with at least four goroutines and open the Bingo Concurrency view.
2. Select goroutine `1` in the spawn tree.
3. Type a filter query in the search field that matches goroutines `2`, `3`, and `4` but not `1`, so the visible node list is `[2, 3, 4]` while `selectedGoroutine` stays `1`.
4. Focus the tree viewport and press `ArrowUp` (or `ArrowLeft`).
5. Observe that goroutine `3` is selected and focused. The expected target is goroutine `4`, the last visible node. Goroutine `4` is skipped entirely.
6. Press `ArrowDown` (or `ArrowRight`) from the same hidden-selection state and observe that goroutine `2` is selected, which is correct — only the reverse direction is wrong.

With exactly two visible nodes the defect is more visible still: reverse selects the first node rather than the last, so reverse and forward navigation produce the identical result.

**Expected behavior**
When the current selection is not present in the visible node list, keyboard navigation should enter the list from the corresponding end: forward directions (`ArrowDown`/`ArrowRight`) select the first visible node, and reverse directions (`ArrowUp`/`ArrowLeft`) select the last visible node. When the selection is visible, the existing wrap-around behavior should be unchanged in both directions. Ancestor-connecting filter behavior, the empty-result early return, and the tree/viewport event handlers should all remain intact.

**Screenshots**
N/A. The problem is reproducible deterministically through DOM keyboard events against the rendered tree.

 - Architecture: darwin/arm64 and linux/amd64 packages
 - Version: VS Code extension `0.3.2`, PR #171 head `a46c1155a8900ee892ee0abeeeaa7b97725a6319`

**Additional context**
Priority: **P3**. The tree remains usable and every node is still reachable with repeated keypresses or a pointer, but reverse keyboard traversal silently skips a node and is inconsistent with forward traversal from the same state, which is a real accessibility defect for keyboard-only users.

This is a separate layer stacked on draft PR #171 (itself stacked on #134). #134 removed dead selector scans from the arrow handler and #171 fixed focus restoration across rerenders; neither changed the index arithmetic. Because this changes shipped webview behavior, the extension package version should be bumped from `0.3.2` to `0.3.3` in `package.json`, the lockfile, the manifest test, the package verifier, the extension README, and `AGENTS.md`.

xsachax:
Fixes #185. Stacked on #171 (base `xsachax-fix-webview-focus-handling`), which is itself stacked on #134. Review only the top commit; the base PRs carry the rest.

## Problem

The concurrency tree keyboard handler in `editors/vscode/src/webviewApp.ts` resolved its position with `findIndex` over the visible nodes and advanced with a single modulo expression:

```ts
const next = visible[(current + direction + visible.length) % visible.length];
```

Filtering narrows the rendered tree but preserves `selectedGoroutine`, so a filtered-out selection makes `findIndex` return `-1`. The modulo then treats that sentinel as a real index one step before the list:

| Direction | Expression | Result |
| --- | --- | --- |
| Forward (`ArrowDown`/`ArrowRight`) | `(-1 + 1 + len) % len` | `0` — first visible node, correct by coincidence |
| Reverse (`ArrowUp`/`ArrowLeft`) | `(-1 - 1 + len) % len` | `len - 2` — **skips the last visible node** |

With visible `[2, 3, 4]` and hidden selection `1`, `ArrowUp` selected g3 and g4 was unreachable in one keypress. With two visible nodes reverse and forward returned the same node.

## Fix

Special-case the absent index so navigation enters the visible list from the matching end — forward at the first node, reverse at the last. A visible selection keeps ordinary wrap in both directions. Ancestor-connecting filter behavior, the empty-result early return, and the tree/viewport handlers are untouched.

## Verification

Reverting only the six-line source change against the new tests reproduces the reported symptom exactly (`ArrowUp from a hidden selection should select g4` → `actual id: 3`), and two of the four new specs fail; restoring it turns them green.

New DOM specs in `test/webviewDom.test.ts`:

- all four arrows against a hidden selection with three visible nodes, asserting both the posted `selectGoroutine` message and `document.activeElement`
- two visible nodes entered from opposite ends
- one visible node reachable from either end
- regression: ordinary wrap in both directions for every position while the selection stays visible

`npm --prefix editors/vscode run check` passes — lint, typecheck, 126 unit/DOM tests, bundle, and package list. The pinned Electron integration test (`npm run test:integration`) passes on three consecutive runs; its single failure was a cold-start flake on the first launch of a freshly downloaded VS Code, reproduced as passing on unmodified base source in the same environment.

## Version

This changes shipped webview behavior, so the extension package moves `0.3.2` → `0.3.3` consistently across `package.json`, `package-lock.json`, `test/manifest.test.ts`, `scripts/verify-package.mjs`, the extension README, and the `AGENTS.md` version pin. The AGENTS.md webview invariant paragraph now records the filtered-navigation rule.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -919,7 +919,7 @@ target metadata, architecture, mode, and entitlements.
 The extension package version is the installed-runtime upgrade boundary:
 material shipped behavior changes must bump both `package.json` and the lockfile
 or VS Code can retain an older bundle under the same identity. The manifest test
-and package verifier pin the current version (**0.3.2**) in source and VSIX
+and package verifier pin the current version (**0.3.3**) in source and VSIX
 metadata.
 The root Run and Debug dropdown exposes exactly two `"type":"bingo"` choices:
 launch one of five progressive examples through a `pickString`, and join a
@@ -1010,8 +1010,13 @@ the 500-node rendering cap, re-lays out each match with at most four ancestors,
 and resets fit so a deep or previously capped match cannot remain invisible.
 Empty results keep Fit/zoom callbacks safe even without an SVG scene. SVG
 treeitems carry `aria-level`, sibling position/size, selection, and parent
-context; arrow navigation moves DOM focus with selection. Full-root model
-rerenders preserve focus through bounded literal identities for the session
+context; arrow navigation moves DOM focus with selection. A filter can hide the
+preserved selection, so navigation special-cases the absent index instead of
+letting `-1` flow into the wrap modulo, which lands on `len-2` in reverse and
+skips the last visible node: forward arrows enter the visible list at its first
+node and reverse arrows at its last, while a visible selection keeps ordinary
+wrap in both directions. Full-root model rerenders preserve focus through
+bounded literal identities for the session
 selector and controls, the two static element IDs, and numeric goroutine IDs,
 but only when `document.hasFocus()` was true before replacement; a blurred
 webview must never reclaim focus from the editor, and an identity absent from
--- a/editors/vscode/README.md
+++ b/editors/vscode/README.md
@@ -21,8 +21,10 @@ just vscode-install
 
 This builds, verifies, and installs `dist/bingo-<platform>.vsix`. The graphical
 concurrency view debuted in 0.3.0; 0.3.1 added capability-safe managed-server
-reuse, and **0.3.2** preserves keyboard focus across telemetry rerenders without
-pulling focus back from the editor. Rerun the command to update, then run
+reuse, 0.3.2 preserves keyboard focus across telemetry rerenders without
+pulling focus back from the editor, and **0.3.3** enters a filtered spawn tree
+from the matching end so reverse arrow navigation reaches the last visible
+goroutine. Rerun the command to update, then run
 **Developer: Reload Window** once so the active extension host loads the new
 bundle. Package without installing with `just vscode-package`. Uninstall with:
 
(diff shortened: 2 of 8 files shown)


xsachax:

PR #192, opened 2026-08-09
fix(client): stop correlating dual-purpose goroutine snapshots

### LINKED ISSUE #187: [client] Auto-pushed goroutine snapshots satisfy the synchronous GoroutineSnapshot call, and the on-demand query consumes lifecycle deltas
**Describe the bug**
**Priority: P2.** `EventGoroutineSnapshot` is the only dual-purpose frame in the wire protocol: it is both an unsolicited server push (auto-emitted on the launch/attach entry stop, every breakpoint hit, and every pause) *and* the answer to `CmdGoroutineSnapshot`. Two semantic defects follow from treating it as a synchronous reply.

**1. The SDK correlates an unsolicited push as a request reply.**
`wsClient.GoroutineSnapshot()` (`pkg/client/ws.go:448-462`) uses `sendAndWait(cmd, protocol.EventGoroutineSnapshot)`, and `routeToPending` matches purely by event kind. Nothing distinguishes an automatic push from this client's query reply, so whichever `EventGoroutineSnapshot` lands first satisfies the waiter.

**2. The on-demand query mutates the automatic stream's lifecycle baseline.**
`engine.GoroutineSnapshot()` (`internal/debugger/engine.go:497-507`) calls `e.goroutineSnapshot()`, which calls `diffGoids` and adopts the new live set into `e.prevGoids` (`internal/debugger/goroutines.go:425-479`) — exactly what the automatic stop snapshots do. A read-only refresh therefore consumes the created/exited deltas that the *next* automatic snapshot should have reported. `prevGoids` is the only lifecycle memory, so those deltas are unrecoverable: the following automatic snapshot diffs against the query, not against the previous stop. The existing e2e concurrency spec already works around this hazard by refusing to call the on-demand path between stops (`test/integration/debugger_e2e_common_test.go:919-923`).

**To Reproduce**
Checked out at `e1d07e7` (branch `xsachax-fix-client-reply-debt`, PR #148, which is the prerequisite reply-debt fix — not a duplicate). Both interleavings are deterministic against the existing `pkg/client` loopback WebSocket harness (`newLoopbackClient`).

*Interleaving A — an automatic push satisfies the synchronous call (pre-existing, independent of #148):*
1. Call `client.GoroutineSnapshot()`; the server reads `CmdGoroutineSnapshot` and withholds its reply.
2. The tracee hits a breakpoint (or is paused), so the hub broadcasts the automatic `EventGoroutineSnapshot` for that stop (`current: 999`, `created: [999]`).
3. Observed:

   ```text
   BUG: sync call returned the unsolicited auto push: current=999 created=[999]
   BUG: the actual query reply arrived on Events() instead: current=1
   ```

   The call returns a snapshot the caller never requested — taken at a different stop — and the real query reply falls through to `Events()`. Callers that render the auto stream (`cmd/wsmon`) double-apply one snapshot and lose the other's ordering.

*Interleaving B — reply debt indefinitely swallows an unrelated automatic push (#148 interaction):*
1. Call `client.GoroutineSnapshot()` and let it time out. Per #148 the request is correctly retired as ordered reply debt matching on `EventGoroutineSnapshot`.
2. The tracee later hits a breakpoint; the hub pushes that stop's automatic `EventGoroutineSnapshot` (`current: 7`, `created: [7]`), followed by an unrelated `EventOutput`.
3. Observed:

   ```text
   first call timed out as expected: timeout waiting for GoroutineSnapshot response
   BUG: the automatic snapshot was swallowed by reply debt; first delivered event was Output
   ```

   The debt claims the automatic push and consumes it without a waiter. The observer never sees that stop's snapshot, and because the engine already advanced `prevGoids` for it, its created/exited deltas are permanently lost. #148 is correct for true confirmations; the defect is that a dual-purpose event was ever listed as one.

*Interleaving C — a manual refresh eats the next stop's deltas:*
1. Suspend at breakpoint N; the automatic snapshot reports deltas relative to stop N−1 and adopts the live set.
2. Send `CmdGoroutineSnapshot` (e.g. `snapshot` in `cmd/cli`, or a UI Refresh). The engine diffs and re-adopts.
3. Continue to breakpoint N+1. Its automatic snapshot now reports deltas relative to the query, not relative to stop N. Any goroutine created and exited between the query and stop N+1 appears in neither snapshot's deltas.

*Interleaving D — the reference client waits forever on a rejected request (`cmd/wsmon -once`):*
1. Start a server, create a session, and leave the tracee **running** (no stop yet).
2. Join it read-only: `go run ./cmd/wsmon -addr localhost:6060 -session <id> -once`.
3. wsmon sends `CmdGoroutineSnapshot` on join. The engine requires a suspended process, so the hub answers `EventError{Command: "GoroutineSnapshot", Message: "process not suspended"}` — and no snapshot.
4. Observed: `-once` prints nothing and never exits. The synchronous path blocked on a reply that never arrives; an event-driven port that only watches for `EventGoroutineSnapshot` ignores the `EventError` and waits just as long. Liveness, not just correctness: no output, no diagnostic, no exit.

**Expected behavior**
- `CmdGoroutineSnapshot` is a *request*, not a synchronous RPC. The SDK must send it and return; every `EventGoroutineSnapshot` — automatic or requested — must be delivered exactly once, in hub order, on `Events()`. It must never register a pending entry or reply debt, so a timed-out or racing request cannot swallow or be satisfied by an unrelated push.
- The on-demand path must be a pure observation: `Created`/`Exited` empty and `prevGoids` untouched. Only the automatic entry/breakpoint/pause snapshots own the lifecycle baseline and its deltas.
- The degraded-snapshot rule is preserved (an unreadable runtime must not touch `prevGoids`), the one-build-per-stop behavior of `emitBreakpointHit`/`emitPaused` is preserved, and there is no wire payload, event-kind, or protocol-version change.
- The reference clients must migrate with the API, and a waiter must stay live. `cmd/wsmon -once` must terminate whether the request is answered, rejected, or silently dropped. It must NOT do so by treating `EventError{Command: GoroutineSnapshot}` as its answer: that event is broadcast and carries no requester, so it may be another client's rejection — and a valid snapshot can still follow it (another client's request is rejected while the tracee runs; the tracee then reaches a stop and pushes a snapshot to every client). Attributing it by kind is the same defect as attributing the snapshot itself by kind. The rejection must therefore be advisory context, a later snapshot must win, and the wait must be bounded by an explicit timeout that reports the last rejection seen when it expires. That timeout must itself be well-formed: only `0` may mean "wait forever", so a negative duration has to be rejected as a usage error rather than silently disarming the bound.

**Screenshots**
N/A — protocol/semantics defect.

 - OS: Darwin 25.5.0
 - Architecture: arm64
 - Version: Go 1.26.1; bingo base `e1d07e7` (PR #148 head, stacked on `main`)

**Additional context**
The dual-purpose nature of the event is why the id-less protocol cannot correlate it: unlike `EventLocals`/`EventFrames`/`EventBreakpointSet`, an `EventGoroutineSnapshot` can be generated with no request at all. Adding correlation IDs is the general fix and remains deferred; making the command fire-and-forget is the correct bounded repair and removes this event from the correlated set entirely.

The `cmd/wsmon` liveness symptom is filed here rather than separately because it is not an independent bug: it is the reference client's half of this migration. The hang exists in both the old synchronous form and any naive event-driven port, and its correct repair is dictated by the same no-correlation property this issue is about — which is why the fix must be a bounded wait rather than by-kind failure attribution. Splitting it would invite a "just exit on the matching EventError" fix that silently re-adds the correlation being removed.

A search across all open and closed issues and pull requests found no existing report for either the dual-purpose correlation case or the `prevGoids` baseline case. #144/#148 is the prerequisite reply-debt fix for genuine confirmations and is not a duplicate; #142 and #173 are DAP-layer correlation/caching issues at a different stack layer.

xsachax:
Fixes #187.

Restacked directly onto exact `origin/main` `2c294461f74f37b785b5beda60fa77cfc56a26c6`. **Six commits — review all of them**; the first is the structural fix and the rest close review findings against it.

| commit | what |
| --- | --- |
| `b6ee9bc` | fire-and-forget snapshot request; on-demand queries stop owning lifecycle baseline |
| `956e333` | mutation-discriminating E2E gate for baseline ownership |
| `68ac826` | `cmd/wsmon -once`: snapshot rejections become advisory and waiting becomes bounded |
| `84adbc7` | reject a negative `-timeout` instead of silently waiting forever |
| `06046d1` | pin that only a zero `-timeout` is indefinite |
| `f5353f7` | print custom usage on validation failures |

## The fix

- `EventGoroutineSnapshot` is dual-purpose: it is pushed automatically on entry/breakpoint/pause and also answers `CmdGoroutineSnapshot`. It therefore cannot be correlated safely to a request by kind.
- Synchronous `Client.GoroutineSnapshot()` is replaced by fire-and-forget `Client.RequestGoroutineSnapshot() error`. It returns after the command reaches the wire, registers no pending waiter or reply debt, and every snapshot is delivered exactly once through `Events()`.
- The snapshot-specific timeout-debt carve-out is removed. Every command still served by `sendAndWait` now has a genuine request-only confirmation and follows the uniform reply-debt rules from #148.
- On-demand queries return the live picture with empty `Created`/`Exited` and never advance `prevGoids`. Only automatic entry/breakpoint/pause snapshots diff and adopt the lifecycle baseline. Degraded snapshots still leave the baseline untouched, and breakpoint/pause stops still build once and reuse that snapshot.
- Reference clients are migrated. `cmd/cli` arms one full render for `snap`; `cmd/wsmon` treats a broadcast snapshot rejection as advisory because it cannot identify the requester.
- **No payload shape, event kind, command kind, or protocol version change.** `internal/hub`, `internal/dap`, and the VS Code extension are untouched.

## `cmd/wsmon -once`

- A snapshot rejection is advisory; a later snapshot always wins.
- `-once` uses `-timeout` (default `30s`) to bound a rejected/no-response request.
- `-timeout 0` is the only indefinite mode and creates no timer.
- Negative durations are rejected with exit 2 and the custom usage text.
- Timeout/closed-stream errors include the last advisory rejection when present.
- Live mode never arms the deadline and keeps observing after rejections.

## Preservation proof

`git range-diff 8dd7593..257793d 2c29446..f5353f7` maps all six commits in order. Commits 2–6 have identical stable patch IDs. Commit 1 had conflicts only in `AGENTS.md` and `internal/debugger/export_test.go`, where current-main documentation and debugger test helpers were retained additively; excluding those two additive files, the old and restacked functional patches have the identical stable patch ID `d227f9b72308e2aa3060ce87e6cb5e7c200cfcc9` and identical stats. Current-main changes from #148/#193 and later debugger/hub/DAP fixes remain present.

## Validation at exact published head `f5353f745b24da00cc28d0ab84fa04d11b96569b`

- `gofmt`, `go tool goimports`, and `git diff --check` clean
- targeted tagged race tests: `pkg/client`, `cmd/cli`, `cmd/wsmon`, `internal/debugger`
- signed targeted Darwin `concurrency` E2E: 2/2
- `go vet -tags 'e2e bingonative' ./test/integration`
- Linux/amd64-context `golangci-lint`: 0 issues
- Linux/amd64 cross-builds for bingo/wsmon and relevant client/CLI/wsmon/debugger/E2E test binaries
- full signed `just e2e-darwin`: **23/23** (current main has one more spec than the earlier 22/22 baseline)
- independent code review: no high-confidence findings

The PR remains **OPEN/DRAFT**.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -302,6 +302,23 @@ automatic — is delivered exactly once on `Events()`, where a client treats the
 uniformly as telemetry. It must stay out of the correlated set: never route it
 through `sendAndWait`. See issue #187.
 
+Two consequences a caller must handle, both inherent to an id-less broadcast
+protocol rather than to this change:
+
+- **Delivery is best-effort.** The answer rides the shared `Events()` buffer,
+  which `readPump` drops from when the consumer stops draining, and a rejected
+  request answers with `EventError`(`CmdGoroutineSnapshot`) instead of a
+  snapshot. Anything that blocks waiting for a snapshot must handle both — see
+  `cmd/wsmon`'s `-once`, which fails fast on a rejection or a closed stream
+  rather than waiting forever.
+- **The answer is broadcast.** A query is fanned out to every client on the
+  session like any other event, and it carries empty `Created`/`Exited`. An
+  observer that renders "the latest snapshot's deltas" (as `cmd/wsmon`'s
+  lifecycle panel does) therefore blanks that panel until the next automatic
+  stop; an append-only timeline (the VS Code view) is unaffected. Making a query
+  addressable to its requester needs wire-level correlation and is deliberately
+  out of scope here — a candidate for a telemetry 1.3.
+
 `syncMu` serializes synchronous commands, but a timeout does **not** cancel the
 command already sent to the server. The pending queue therefore retains a
 timed-out request as retired reply debt until its matching confirmation or
@@ -1657,6 +1674,16 @@ consumed the pending deltas — the following automatic snapshot diffed against
 the query, so goroutines created and exited in between appeared in neither
 (issue #187). Do not add a delta-bearing query path back: the wire payload is
 unchanged, and a client that wants deltas must read the automatic stream.
+
+The regression gate is the `concurrency`-labelled `declareBaselineOwnershipSpec`
+E2E, and it is deliberately shaped: a query at an ordinary breakpoint stop sees
+exactly the live set that stop's automatic snapshot already adopted, so adopting
+again would be a no-op and prove nothing. The spec instead **steps over a `go`
+statement** — steps are the one suspend that carries no automatic snapshot — so
+the query observes a goroutine the baseline has never seen, then asserts the
+NEXT automatic snapshot still reports it as created. It fails if the query is
+made lifecycle-tracking or the automatic path non-tracking; the unit tests
+around `snapshotFrom` pin the seam's semantics but cannot catch a rewiring.
 
 **Graceful fallback.** Every read is best-effort. `resolveGoLayout` marks the
 layout invalid if any required `g`/`gobuf`/`stack`/`m` offset is missing, and any
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -308,9 +308,13 @@ protocol rather than to this change:
 - **Delivery is best-effort.** The answer rides the shared `Events()` buffer,
   which `readPump` drops from when the consumer stops draining, and a rejected
   request answers with `EventError`(`CmdGoroutineSnapshot`) instead of a
-  snapshot. Anything that blocks waiting for a snapshot must handle both — see
-  `cmd/wsmon`'s `-once`, which fails fast on a rejection or a closed stream
-  rather than waiting forever.
+  snapshot. Anything that waits for a snapshot must handle both — and must not
+  treat that error as *its* answer: `EventError` is broadcast and carries no
+  requester, so it may be another client's rejection with a valid snapshot still
+  to come. Correlating it by kind is the same mistake as correlating the
+  snapshot itself. `cmd/wsmon`'s `-once` shows the intended shape: the rejection
+  is advisory, a later snapshot wins, and a `-timeout` deadline (not the error)
+  is what bounds the wait.
 - **The answer is broadcast.** A query is fanned out to every client on the
   session like any other event, and it carries empty `Created`/`Exited`. An
   observer that renders "the latest snapshot's deltas" (as `cmd/wsmon`'s
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -276,12 +276,31 @@ In [pkg/client](pkg/client/), the `Client` interface splits methods by what
 they wait for:
 
 - **Synchronous** (`Restart`, `SetBreakpoint`, `ClearBreakpoint`, `Locals`,
-  `Evaluate`, `StackFrames`, `Goroutines`, `GoroutineSnapshot`): block until
-  the matching confirmation event (or `EventError` for the same command kind)
-  arrives. Implemented via `sendAndWait` in [pkg/client/ws.go](pkg/client/ws.go).
-- **Fire-and-forget** (`Launch`, `Attach`, `Kill`, `Continue`, `Step*`, `Pause`):
-  return as soon as the command is on the wire. Results arrive asynchronously
-  on the `Events()` channel.
+  `Evaluate`, `StackFrames`, `Goroutines`): block until the matching
+  confirmation event (or `EventError` for the same command kind) arrives.
+  Implemented via `sendAndWait` in [pkg/client/ws.go](pkg/client/ws.go).
+- **Fire-and-forget** (`Launch`, `Attach`, `Kill`, `Continue`, `Step*`, `Pause`,
+  `RequestGoroutineSnapshot`): return as soon as the command is on the wire.
+  Results arrive asynchronously on the `Events()` channel.
+
+`RequestGoroutineSnapshot` is fire-and-forget **for a correctness reason, not
+convenience**: `EventGoroutineSnapshot` is the only dual-purpose frame in the
+protocol — the server also pushes it unsolicited on every entry/breakpoint/pause
+stop — so it can never be correlated to a request by kind. A synchronous
+`GoroutineSnapshot()` has no sound placement in the reply-debt model below, in
+either direction. *Retiring* a timed-out snapshot request as debt lets that debt
+silently swallow the next automatic push, permanently losing that stop's
+lifecycle deltas. *Discarding* it instead re-opens #144 for this kind: a stale
+reply — or any automatic push — then satisfies a later same-kind call. Even
+before any timeout, a plain in-flight call is satisfied by whichever snapshot
+lands first, which is usually a stop's automatic push. All three follow from one
+root: a frame that can exist with **no request** cannot be matched to one.
+
+So the fix is to remove the waiter, not to place it better. The request registers
+**no pending entry and no reply debt**, and every snapshot — requested or
+automatic — is delivered exactly once on `Events()`, where a client treats them
+uniformly as telemetry. It must stay out of the correlated set: never route it
+through `sendAndWait`. See issue #187.
 
 `syncMu` serializes synchronous commands, but a timeout does **not** cancel the
 command already sent to the server. The pending queue therefore retains a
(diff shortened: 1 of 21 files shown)


xsachax:
Local darwin/arm64 E2E verification for the `darwin-e2e-verified` gate.

Ran `just e2e-darwin` (codesigned with `entitlements.plist`, real Mach-exception backend, Apple Silicon) against `991678c5f753100be03276611934fdb585e2f3d5`:

```
Ran 23 of 23 Specs in 257.732 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

The gate tripped because this head edits `test/integration/debugger_e2e_darwin_arm64_test.go` — one line wiring the new `declareBaselineOwnershipSpec` into the darwin container. No darwin backend code changed.

The new `concurrency` spec was additionally run 5× in isolation (all pass) and proven mutation-discriminating: reverting `engine.GoroutineSnapshot` to the lifecycle-tracking path fails it, and making the automatic path non-tracking fails it too.

xsachax:
Re-verification for the `darwin-e2e-verified` gate after `e7ddf93` (the label is auto-removed on every push).

`just e2e-darwin` against `e7ddf936a5f1304565b4395dc4c318a1d02d35b6`:

```
Ran 23 of 23 Specs in 255.719 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

This head changes only `cmd/wsmon`, `pkg/client` docs, AGENTS.md, and `docs/ConcurrencyTelemetry.md`; no darwin backend or E2E spec code. The gate still applies because the PR as a whole edits `test/integration/debugger_e2e_darwin_arm64_test.go` (one line wiring `declareBaselineOwnershipSpec` into the darwin container).

xsachax:
Re-verification for the `darwin-e2e-verified` gate after `0b9fb83` (the label is auto-removed on every push).

`just e2e-darwin` against `0b9fb833a7d8e2ca9c5711f8b43a0708b6b710e8`:

```
Ran 23 of 23 Specs in 266.371 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

This head changes only `cmd/wsmon` (flag validation) and `docs/ConcurrencyTelemetry.md`; no darwin backend or E2E spec code. The gate still applies because the PR as a whole edits `test/integration/debugger_e2e_darwin_arm64_test.go` (one line wiring `declareBaselineOwnershipSpec` into the darwin container).

xsachax:
Confirmation re-run of the Darwin verification against the final head, independent of the earlier run.

`just e2e-darwin` at `0b9fb833a7d8e2ca9c5711f8b43a0708b6b710e8` (clean tree, no commits since):

```
Ran 23 of 23 Specs in 259.394 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

Second independent pass on this head (the first was 266.371s, same 23/23). No push has occurred since, so the `darwin-e2e-verified` label applied after that first run still covers exactly these commits and the gate is green.

xsachax:
Darwin verification for the final head `d191810` (the previous label was cleared by this push, as designed).

`just e2e-darwin` against `d19181007de34569f963cd0ec2b07acbfddc72b3`, clean tree:

```
Ran 23 of 23 Specs in 259.679 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

This head adds `cmd/wsmon` tests only; no darwin backend or E2E spec code changed. The gate applies because the PR as a whole edits `test/integration/debugger_e2e_darwin_arm64_test.go` (one line wiring `declareBaselineOwnershipSpec` into the darwin container).

xsachax:
Darwin verification for the final head `43c15d8` (the push cleared the previous label, as designed).

`just e2e-darwin` against `43c15d833e296d7742c0709a795143e46c0932fd`, clean tree:

```
Ran 23 of 23 Specs in 258.171 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

This head touches `cmd/wsmon` only (usage reporting on validation failure plus its tests); no darwin backend or E2E spec code changed. The gate applies because the PR as a whole edits `test/integration/debugger_e2e_darwin_arm64_test.go` (one line wiring `declareBaselineOwnershipSpec` into the darwin container).

xsachax:
Darwin verification for the final head `9534026` (the force-push cleared the previous label, as designed).

`just e2e-darwin` against `95340260bba324eb02d5a56aa1c04b24a52ccbb7`, clean tree:

```
Ran 23 of 23 Specs in 257.832 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

`9534026` amends `43c15d8` with the errcheck fix golangci-lint flagged (`_, _ = fmt.Fprintf(fs.Output(), …)`, matching the existing style in `bindFlags`). Behavior is identical — verified on a built binary — and `golangci-lint run --build-tags bingonative` is now clean locally.

xsachax:
Darwin verification for the frozen head `257793d`, now rebased onto the accepted #148 base `8dd7593`.

`just e2e-darwin` (codesigned with `entitlements.plist`, real Mach-exception backend, Apple Silicon), clean tree:

```
Ran 23 of 23 Specs in 256.542 seconds
SUCCESS! -- 23 Passed | 0 Failed | 0 Pending | 0 Skipped
```

This supersedes an earlier local run on this same head in which the `churn` spec failed with `StepOver #N unexpected Error: {"message":"stop the world failed"}`. That failure was **machine load, not a regression** — it reproduced identically on a clean worktree at the base `8dd7593` with none of this PR's code, and `churn` passed on Linux CI at this head throughout. The failing runs were at load average ~22 on a 10-core box; this passing run was at ~9.6. Nothing in this PR touches stepping or Mach code.

Also green on this head: `go test -tags bingonative -race -count=1 ./...` (all 12 packages), `go vet -tags bingonative ./...`, `gofmt`/`goimports`, and `golangci-lint run --build-tags bingonative` (0 issues).

No code changed since the previous push — the head is byte-identical, so this attests the exact commit under review.

PR #193, opened 2026-08-09
fix(hub): dispose of an abandoned restart replacement

### LINKED ISSUE #188: [hub] Failed Restart relaunch leaks the replacement debugger
**Describe the bug**

**Priority: P2.** Confirmed on `origin/main` at `0509b53` and still present on the #164 branch (`xsachax-fix-hub-debugger-ownership`, `cd77be7`).

`Hub.handleRestart` kills the old debugger, constructs a replacement via the `newDebugger` factory, and calls `Launch` on it. If that relaunch **fails**, the replacement is dropped on the floor without `Kill()`. It was never installed in `h.dbg`, so no other owner can ever reach it:

- `shutdown()` only kills the debugger stored in `h.dbg` — the failed replacement was never stored there.
- `Run` never selects on its `Events()` channel for the same reason.
- The client only sees an `EventError`; nothing tears the instance down.

`debugger.New` is **not** a cheap value. It constructs live OS resources before `Launch` is ever called:

- `newEngine` (`internal/debugger/engine.go:170-186`) starts `go e.loop()` immediately.
- `engine.loop` (`internal/debugger/engine.go:509-519`) calls `runtime.LockOSThread()` and never unlocks. Its `defer` is the **only** place that closes `done`/`events` and calls `closeTracer()`.
- On linux, `newBackend()` (`internal/debugger/backend_linux_amd64.go:17-19`) constructs a `tracerThread` — a second goroutine that also pins an OS thread — at construction time, before any launch.
- `startTracedProcess` can also fail *after* forking (`Wait4`, `PTRACE_SETOPTIONS`), so a partially started tracee can be involved.

Only `Kill()` drives the engine loop to `stateExited` and lets that `defer` run. So each abandoned replacement permanently leaks one engine goroutine + locked OS thread and, on linux, one tracer goroutine + locked OS thread. Repeated failed Restarts accumulate them for the lifetime of the server process; the session itself remains usable (state returns to `idle`), so a client can retry the failing Restart in a loop.

**Locations**

`origin/main` @ `0509b53`, `internal/hub/hub.go:547-553`:

```go
newDbg := h.newDebugger()
if err := newDbg.Launch(program, args, env); err != nil {
    h.broadcastError(cmd.Kind, fmt.Errorf("restart: relaunch failed: %w", err))
    h.transitionState(protocol.StateIdle)
    return                      // newDbg dropped, never killed
}
h.setDbg(newDbg)
```

#164 branch @ `cd77be7`, `internal/hub/hub.go:612-628` — #164 correctly discards the candidate on the constructor gap (`:613-616`) and on install rejection (`:625-628`), but both relaunch-failure exits still return without disposal:

```go
if err := newDbg.Launch(program, args, env); err != nil {
    if h.isClosing() {
        return                  // leak: shutdown snapshotted h.dbg == nil
    }
    h.broadcastError(cmd.Kind, fmt.Errorf("restart: relaunch failed: %w", err))
    h.transitionState(protocol.StateIdle)
    return                      // leak
}
```

By contrast the initial Launch/Attach path in `executeCommand` (`internal/hub/hub.go:453-468` on `cd77be7`) *does* `discardDebugger(dbg)` when its candidate's dispatch fails. Restart is the asymmetric path.

**To Reproduce**

Deterministic, no sleeps, using the existing `internal/hub` fakes. The key harness detail is that the factory must return a **distinct** fake per call, otherwise a shared fake's call log cannot distinguish which instance was killed:

1. Start a managed hub whose factory hands out a fresh `fakeDebugger` per call; make every call after the first return a `launchErr`.
2. Connect one client, drain the welcome event, and complete an initial `Launch` (factory instance 0).
3. Inject `CmdRestart`.
4. Wait for the resulting `EventError`.
5. Assert the second factory instance's recorded calls.

```go
It("kills the replacement whose relaunch failed", func() {
    factory := newRecordingFactory(func(n int, fd *fakeDebugger) {
        if n > 0 {
            fd.launchErr = errors.New("relaunch boom")
        }
    })
    _, conn, cancel := newManagedHub(factory.create)
    defer cancel()
    launchManagedFactory(conn, factory, "myapp")

    conn.inject(mustCommand(protocol.CmdRestart, protocol.RestartPayload{}))
    waitForEventKind(conn, protocol.EventError, nil)

    Eventually(factory.killCounter(1), "500ms", "10ms").Should(Equal(1))
    Expect(factory.instance(1).recordedCalls()).To(Equal([]string{"Launch", "Kill"}))
})
```

**Observed** (`go test -tags bingonative ./internal/hub/...`):

```
[FAILED] Timed out after 0.500s.
the abandoned replacement must be killed exactly once
Expected
    <int>: 0
to equal
    <int>: 1
```

The replacement's call log is `["Launch"]` — never `Kill`. The `running → idle` transition and the `EventError` both fire correctly; the disposal is the only thing missing.

**Expected behavior**

A replacement debugger is caller-owned from construction until `installDebugger` accepts it. Every exit from that window — relaunch failure, install rejection, or a shutdown race — must dispose of that exact candidate exactly once, outside `dbgMu`, and must never kill a newer or currently installed debugger. A cleanup failure should be logged at the hub (the owning top level per `docs/ErrorHandling.md`) without replacing the original relaunch error that the client is told about.

The existing behavior on failure must be preserved: the old debugger is still killed, the session still transitions to `idle`, `EventError` still carries the wrapped `restart: relaunch failed:` cause, `lastLaunch`/`restartBreakpoints` are retained so a retry works, and a successful restart is unaffected.

**Screenshots**

N/A — deterministic lifecycle defect.

 - OS: Linux and macOS
 - Architecture: amd64 and arm64
 - Version: `origin/main` at `0509b53`; also `cd77be7` (#164 branch)

**Additional context**

Duplicate search over all open/closed issues and PRs for `restart`, `relaunch`, `leak`, `ownership`, and `goroutine leak debugger` found no existing report.

Related:

- #157 / #164 fix the **shutdown-vs-install** ownership race and introduce `installDebugger` / `detachDebugger` / `discardDebugger`. They deliberately scope to that race; this failed-launch disposal gap is pre-existing and disclosed, not caused, by that work. This issue is the natural follow-up and stacks on it.
- #86 introduced the hub-level Restart sequence containing the gap.
- #174 concerns a *rejected* Restart abandoning the suspended wait loop — a different failure mode (hub loop state) on the same command.

xsachax:
Fixes #188.

Restacked as a single #193-specific commit onto `main` at `5c5268192e30affad72ef2a181fb28083662406d`. The stacked #164 ownership work and #180 suspended-loop fix are already on `main`; this branch replays only the accepted replacement-engine disposal change, tests, and ownership/restart documentation.

## Problem

`handleRestart` kills the old debugger, constructs a replacement, and calls `Launch`. When relaunch fails, the replacement is still caller-owned: it is not installed in `h.dbg`, shutdown cannot see it, and `Run` never observes its events. Dropping it therefore strands the engine goroutine and locked OS thread (plus Linux's tracer thread), accumulating resources across restart retries.

## Fix

Register a deferred discard immediately after replacement construction and clear it only at the successful `installDebugger` ownership-transfer point. This structurally covers relaunch failure, shutdown winning before install, and future early returns without killing a successfully installed debugger.

`discardDebugger` is now a `Hub` method and logs cleanup failures at the owning top level without replacing the original client-visible relaunch error.

Breakpoint reinstall failures remain non-fatal: after install, the debugger is hub-owned and the running replacement stays alive while unresolved locations are reported as discarded breakpoints.

## Coverage

The hub regression suite covers failed relaunch disposal, partially acquired resource release, repeated failures, retryability, cleanup errors, successful ownership transfer, shutdown races, saved breakpoints, and partial/total breakpoint reinstall failure. Later hub/DAP/debugger behavior from current `main` is preserved.

## Validation

Exact head: `d0e471b1ff875e3a9c389371883ccee37fbcaef3`

- `gofmt` + `go tool goimports`
- `go test -tags bingonative -race -count=1 ./internal/hub/... ./internal/dap/... ./internal/debugger/...`
- `go test -tags bingonative -count=1 ./...`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 golangci-lint run ./...` (`0 issues`)
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build ./...`
- `just e2e-darwin`: **22/22 passed** on this exact head
- Independent final diff review: no findings

The PR remains draft and must not be merged until exact-head CI is green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1656,6 +1656,29 @@ the lock and perform the idempotent `Kill` outside it. State transitions also
 check `closing` under `dbgMu`, preventing a command already in flight from
 resurrecting a session after registry teardown begins.
 
+**A candidate must be disposed of on EVERY exit, not just the shutdown race.**
+Being caller-owned cuts both ways: while a candidate sits outside `h.dbg`,
+shutdown's snapshot cannot see it and `Run` never selects on its events, so if
+the command path abandons it *nothing else ever will*. `debugger.New` is not a
+cheap value — `newEngine` starts `go e.loop()` immediately, that loop
+`runtime.LockOSThread()`s, and its `defer` is the only thing that closes
+`done`/`events` and calls `closeTracer()`; on linux `newBackend` also spins up a
+tracer thread at construction. **Only `Kill` drives that loop to exit**, so a
+dropped candidate permanently strands a goroutine plus a locked OS thread per
+attempt (issue #188: repeated failed Restarts accumulated them). `handleRestart`
+therefore registers a deferred `discardDebugger` the moment the candidate
+exists and clears it at the single ownership-transfer point — immediately after
+`installDebugger` accepts it — so relaunch failure, install rejection, and any
+future early return all dispose exactly once, and a successful restart's now
+hub-owned debugger is never killed by the caller. Disposal targets the captured
+candidate by identity, so it can never kill a newer or currently installed
+debugger.
+
+`discardDebugger` is a `Hub` method because the hub is the last owner of a
+discarded debugger: a failing `Kill` is logged there (the owning top level, per
+[docs/ErrorHandling.md](docs/ErrorHandling.md)) rather than returned, so cleanup
+never replaces the original launch error the client is told about.
+
 ## Restart — hub-level, not engine-level
 
 `CmdRestart` (`internal/hub/hub.go` → `handleRestart`) kills the current
@@ -1721,6 +1744,25 @@ see [Suspend/resume protocol](#suspendresume-protocol).
 suspending one — the new process's suspended state (if any, e.g. break-on-
 entry) is reported the normal way via `EventStepped`/`EventBreakpointHit` once
 the relaunched process actually reaches that point.
+
+**Failed relaunch**: the old debugger is already dead by then, so the session
+falls back to `idle` with no debugger installed and broadcasts an `EventError`
+wrapping the cause as `restart: relaunch failed: …`. `lastLaunch` and
+`restartBreakpoints` are deliberately **kept**, so a retry Restart (or a fresh
+Launch) still works. The replacement that failed to launch is killed by
+`handleRestart` itself — see the candidate-disposal rule in
+[Hub debugger ownership](#hub-debugger-ownership--shutdown-is-a-linearization-point).
+
+**A failed breakpoint reinstall is NOT a failed relaunch** — do not extend the
+disposal rule to it. By the time the reinstall loop runs, `installDebugger` has
+already accepted the replacement, so it is hub-owned and the tracee is running;
+a `SetBreakpoint` error is collected as a `DiscardedBreakpoint` and reported in
+`EventRestarted` (mirroring Delve), and the session stays `running`. Killing the
+replacement there would terminate a healthy process over one unresolvable
+`file:line`. Only locations that *did* resolve are carried into
+`restartBreakpoints`. Both directions are pinned by the `Restart relaunch
+failure` and `Restart breakpoint reinstall failure` specs in
+[hub_test.go](internal/hub/hub_test.go).
 
 ## Breakpoint identity — hub-owned logical ids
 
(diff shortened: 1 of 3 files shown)


xsachax:
Temporarily closing and reopening unchanged head to trigger the full pull_request CI matrix after the draft restack.

Issue #111, opened 2026-07-20
flaky: Linux 'E2E kill' hangs — cmd.Wait() deadlocks racing the engine's wait4 reaper


xsachax:
## Symptom

The `E2E kill (kill-while-running)` step of the **Linux amd64 (ptrace)** job flaked on `main` after PR #108 merged ([run 29715609273](https://github.com/bingosuite/bingo/actions/runs/29715609273)): the step hung until the 5-minute test timeout (`panic: test timed out after 5m0s`) even though the identical commit passed on the PR beforehand. Intermittent → a race, not a deterministic break.

## Root cause

`internal/debugger/backend_linux_amd64.go` `killProcess` reaps a launched tracee with `cmd.Wait()`. `engine.Kill` runs `killProcess` **on the engine loop thread**. While the tracee is running there is also an active `waitLoop` blocked in `Wait4(-1, WALL)`. Both are reapers of the same child:

- If `cmd.Wait()` wins the reap → fine (today's happy path).
- If the `waitLoop` wins → `cmd.Wait()`'s pidfd-based `waitid` **never returns** (the child is already reaped), so the engine loop is stuck inside the Kill closure forever, `dispatch` blocks, and nothing tears the engine down. → 5-minute timeout.

The goroutine dump from the failing run confirms it: the engine loop is parked in `os/exec.(*Cmd).Wait → pidfdWait → Waitid` for 4 minutes, the `waitLoop` goroutine is already gone (it reaped and exited), and `engine.Kill` is blocked in `dispatch`.

Darwin is immune: its `waitLoop` blocks on a Mach exception port, not `wait4`, so its `killProcess` `Wait4(pid)` is the sole reaper.

## Fix

Reap with `syscall.Wait4` instead of `cmd.Wait()`. `wait4` returns `ECHILD` immediately for the racing loser, so Kill stays responsive whichever reaper wins. Add a `kill` e2e regression loop (`BINGO_E2E_KILL_ITERS`) so the reap race is exercised deterministically in CI.

Issue #199, opened 2026-08-09
[debugger] Linux sibling breakpoints corrupt an in-flight single-step


xsachax:
**Describe the bug**

At `origin/main` commit `0509b53bfc44da889e152ee20f56ce3663882c65`, the Linux/amd64 debugger can surface a sibling thread's breakpoint while another thread is single-stepping off a software breakpoint. The engine receives that sibling stop while the original trap is temporarily restored to its instruction bytes and `steppingOverBP` is a single global slot.

Two production failures follow:

1. **Distinct breakpoint:** the sibling `StopBreakpoint` replaces `lastBP`; the next resume overwrites `steppingOverBP`, permanently losing the first breakpoint entry after `removeFromTable`. Its trap remains disarmed, both `byID` and `byAddr` forget it, and the original ID can no longer be cleared.
2. **Same-address sibling:** while the trap entry is absent, a second thread stopped at that INT3 is treated as a spurious SIGTRAP. On amd64 its RIP is already `addr+1`; the current spurious path advances it again and resumes it in the middle of the restored instruction. `ContinueProcess` also clears `stepping`/`stepTID`, so the real step completion is subsequently misclassified.

This is a P1 debugger-integrity defect: breakpoints can silently stop firing, step requests can lose their completion event, and the tracee can resume at an invalid instruction boundary.

**Code locations**

- `internal/debugger/backend_linux_amd64.go:429-579` — `Wait4(-1, WALL)` can return any traced TID. `stepping` + `stepTID` classify a foreign breakpoint correctly but return it immediately instead of preserving the active step's atomic state-machine boundary.
- `internal/debugger/backend_linux_amd64.go:597-602` — `ContinueProcess` implicitly targets `lastStopTID`; this is why a deferred stop must be recorded only when actually returned.
- `internal/debugger/engine.go:589-752` — `handleStop` consumes the foreign breakpoint/signal while a step is in flight.
- `internal/debugger/engine.go:1031-1065` — `resumeFromBreakpoint` removes the breakpoint table entry and stores the one in-flight `steppingOverBP` slot, which a later resume can overwrite.
- `internal/debugger/breakpoint.go:48-89,96-114` — table removal/reinstall mechanics used by the proof.

**To Reproduce**

1. On native `linux/amd64`, build a Go target with two locked OS threads repeatedly reaching armed breakpoint locations A and B.
2. Stop at A and resume so bingo restores A's instruction and issues an exact-TID single-step.
[...]

PR #201, opened 2026-08-09
fix(debugger): keep a linux step-over atomic against sibling stops

### LINKED ISSUE #198: [debugger] linux: a sibling thread's breakpoint stop consumed mid-step-over silently orphans the stepped-over breakpoint
**Describe the bug**

On linux/amd64 a software-breakpoint step-over can be destroyed by an unrelated thread, silently losing a user breakpoint and resuming a thread mid-instruction.

Linux ptrace stops are per-thread and bingo does not stop the world, so sibling threads keep running while the engine is `stateSuspended`. No `waitLoop` runs while suspended, so a sibling that hits its own armed `INT3` sits in a tracing-stop with its wait status **pending and unreaped**.

`engine.resumeFromBreakpoint` ([engine.go:1031-1068](../blob/main/internal/debugger/engine.go)) then begins a step-over:

```go
bp := e.lastBP; e.lastBP = nil
e.steppingOverBP = bp            // ONE SLOT
e.bps.removeFromTable(bp)        // leaves byID AND byAddr
WriteMemory(bp.addr, bp.originalBytes)  // INT3 DISARMED in tracee memory
SingleStep(tidA)                 // backend: stepping=true, stepTID=tidA
```

The very next `Wait4(-1, WALL)` can return the **sibling's** already-pending `cause==0` SIGTRAP before the stepping thread's trap. `backend_linux_amd64.go:523` correctly classifies it as `StopBreakpoint` (`tid != stepTID`). But `handleStop`'s `case StopBreakpoint:` ([engine.go:589-636](../blob/main/internal/debugger/engine.go)) has **no `steppingOverBP` guard** — line 600 only *logs* the flag — so it overwrites `lastBP`/`lastBPTID` and emits `BreakpointHit` while `steppingOverBP` still holds the half-stepped entry. `steppingOverBP` is only ever consumed by the `StopSingleStep` (655-670) and `StopSignal` (729-740) branches, neither of which runs. The next `Continue` reassigns `e.steppingOverBP` at engine.go:1034, and the entry is then referenced by nothing: not `byID`, not `byAddr`, not `lastBP`, not `steppingOverBP`.

**Two harms, both observed on real CI:**

1. **Silent, unrecoverable loss of a user breakpoint (High).** It is disarmed in tracee memory and gone from the table: it never fires again, and `ClearBreakpoint` returns `breakpoint N not found`. Tracee memory itself is left consistent.

2. **Mid-instruction resume of the sibling (Critical).** When the sibling trapped at the *same* address currently being stepped over, `bps.atAddr` returns nil (the entry was just removed), so `handleStop` takes the *spurious SIGTRAP* branch (engine.go:601-615). That branch **does not call `rewindToBreakpoint`**. On amd64 its `regs.PC = stop.PC + len(INT3)` write is a no-op (`archRewindPC` subtracted 1, this adds it back), so the thread is resumed with `RIP = addr+1` — one byte into the multi-byte original instruction that was just restored. Arbitrary misdecoded execution.

This is a **shared-engine** gap, not a linux quirk: darwin's default backend papers over it in `Wait` (only the stepping thread's trap can surface while `sobActive`), but its ablation path `BINGO_DARWIN_ATOMIC_STEPOVER=0` (backend_darwin_arm64.go:423-425, :872-886) returns `StopBreakpoint` for a sibling in exactly the same shape.

**To Reproduce**

Proof branch `xsachax-proof-linux-sibling-bp-overlap` (test-only; production code identical to `0509b53`) adds an `overlap`-labelled real-ptrace E2E:

```sh
go test -tags e2e -race -count=1 -v -timeout 900s ./test/integration -ginkgo.v -ginkgo.label-filter=overlap
```

It parks one pinned OS thread on `MARK_A` and four pinned siblings on `MARK_B`, releases them together so they race to their traps inside one running window, verifies via `/proc/<pid>/task/*/status` + `/proc/<pid>/task/<tid>/syscall` that the threads are parked **on** the marker instructions (addresses read from the target's own DWARF; an amd64 INT3 stop leaves RIP at `addr+1`), then resumes and probes both breakpoints after every stop.

Detector: re-issuing `SetBreakpoint` on the stepped-over line. `breakpointTable.set` checks `byAddr` **before** touching memory, so success proves the entry is gone. Corroborated read-only by `ClearBreakpoint` on the original id (consults `byID` before any write).

**Expected behavior**

Both breakpoints remain installed across a step-over regardless of which thread `Wait4` reports next, and no thread is ever resumed at a PC that is not an instruction boundary.

**Observed**

`SetBreakpoint(overlap_target.go:18)` **succeeded** (must return `errBreakpointExists`), and `ClearBreakpoint(id=1)` returned `breakpoint 1 not found`.

Engine's own debug log at the failing stop:

```
DEBUG StopSingleStep  pc=0x4b00c5 steppingOverBP=true
DEBUG breakpoint reinstalled addr=0x4b00be                       <- a healthy step-over
DEBUG StopBreakpoint  pc=0x4b007e found=true  steppingOverBP=false  <- MARK_A hit, lastBP=bpA
DEBUG StopBreakpoint  pc=0x4b00be found=true  steppingOverBP=true   <- OVERLAP: lastBP overwritten
```

And the harm-2 variant (sibling parked on the address being stepped):

```
DEBUG StopBreakpoint pc=0x4b00be found=false steppingOverBP=true
WARN  spurious SIGTRAP — advancing PC past BRK and resuming pc=0x4b00be   (x2)
```

`/proc` at detection showed the stepping thread still in tracing-stop at exactly `MARK_A` with its trap disarmed.

**Desktop**

- OS: Linux (ubuntu-24.04 GitHub-hosted runner, kernel 6.17.0-1020-azure), Go 1.25.5, `ptrace_scope=1`
- Architecture: amd64
- Version: `0509b53`

**Additional context**

Reproduced **4 of 4 attempts** in run [31324476072](https://github.com/bingosuite/bingo/actions/runs/31324476072), and 3 of 4 in [31323712681](https://github.com/bingosuite/bingo/actions/runs/31323712681) / [31323980359](https://github.com/bingosuite/bingo/actions/runs/31323980359) — including an attempt **without `-race`**, so this is not a race-detector timing artifact. Every measured cycle met the non-vacuity precondition (`atMarkerPC` 2 or 4).

Existing E2E coverage misses this because every current target has only one goroutine executing the breakpointed line, so two threads are never parked on armed traps simultaneously.

Two independent adversarial reviews confirmed the causal chain and the detector's soundness. They disagreed on the remedy: an engine-side reinstall in the `StopBreakpoint` branch makes the loss recoverable and (because of its placement before `atAddr`) also repairs harm 2, but it abandons a still-in-flight `PTRACE_SINGLESTEP`, leaving misclassification hazards. The root cause is that `Wait4(-1, WALL)` **consumes** a sibling's status during the step-over window; keeping the step atomic at the backend is the more complete remedy.

xsachax:
Fixes #198.

## The bug

`engine.resumeFromBreakpoint` restores a breakpoint's original bytes, drops the entry from the breakpoint table and single-steps one thread; only the resulting `StopSingleStep` puts the trap back. Linux ptrace stops are per-thread and bingo does not stop the world, so siblings keep running while the engine is suspended and can be sitting in their own unreaped `INT3` stops.

A process-wide `Wait4(-1, WALL)` **consumed** such a status inside that window and handed the engine a `StopBreakpoint` it could not act on. `handleStop`'s `StopBreakpoint` branch overwrote `lastBP`/`lastBPTID` while `steppingOverBP` still held the half-stepped entry, and the next `Continue` overwrote that last reference. Two harms:

1. **Silent, unrecoverable loss of a user breakpoint** — disarmed in memory, gone from `byID`/`byAddr`: never fires again, and `ClearBreakpoint` returns `not found`.
2. **Mid-instruction resume** — when the sibling had parked on the *same* address being stepped, the table lookup missed, so `handleStop` took the spurious-SIGTRAP path, which skips `rewindToBreakpoint`, and the thread was resumed one byte into the restored multi-byte instruction.

This is a **shared-engine** gap, not a linux quirk: darwin's default backend papers over it in `Wait`, but its `BINGO_DARWIN_ATOMIC_STEPOVER=0` ablation path returns exactly the same shape.

## The fix

**Backend — the root cause.** While a single-step is in flight, target `wait4` at the stepping thread instead of any thread (`waitTarget`). Sibling statuses then stay queued in the kernel — nothing is lost — and surface on the next process-wide wait with the trap reinstalled. `ECHILD` from the targeted wait means the stepping thread is gone, so it clears the step state and falls back to `Wait4(-1)`: the waitLoop must remain the sole reaper of every thread's death for `Kill` (#111).

**Engine — defense in depth.** Reinstall an in-flight `steppingOverBP` in the `StopBreakpoint` branch, mirroring the existing `StopSignal` branch. Backends that can still deliver a sibling stop mid-step-over then degrade to a correctly-reported hit instead of a lost breakpoint. Rearming **before** `bps.atAddr` is load-bearing — it is what makes the same-address case resolve to a real hit rather than the spurious-SIGTRAP path — so it must not be reordered below the lookup.

## Evidence: A/B on native hosted CI

Identical `overlap` spec, ubuntu-24.04, real ptrace.

| Branch | Production code | `E2E overlap` |
| --- | --- | --- |
| [`scratch-198-unfixed`](https://github.com/bingosuite/bingo/tree/scratch-198-unfixed) ([run](https://github.com/bingosuite/bingo/actions/runs/31325539248)) | `main` @ `0509b53` | **fails in 2.7s** |
| this branch ([run](https://github.com/bingosuite/bingo/actions/runs/31325537802)) | with the fix | **passes** |

Unfixed control, verbatim:

```
DEBUG StopBreakpoint pc=0x4b00be found=false steppingOverBP=true
WARN  spurious SIGTRAP — advancing PC past BRK and resuming pc=0x4b00be   (x2)
DEBUG StopBreakpoint pc=0x4b007e found=true  steppingOverBP=true
[FAILED] breakpoint B (MARK_B) was silently orphaned by a sibling stop during a step-over
  Probe: SetBreakpoint(overlap_target.go:25) SUCCEEDED.
  Read-only corroboration: ClearBreakpoint(id=2) returned "breakpoint 2 not found".
  armedA=true armedB=false   atMarkerPC(max)=4
```

With the fix, all four jobs are green and **every existing spec still passes** (basic, churn, pause, stepping, inspect, breakpoints, kill, exit, attach, concurrency, fullstack, restart, dap; darwin compiles + codesigns). The new spec reports `stops=150 hitsA=30 hitsB=120 atMarkerPC(max)=4 overlapReady=120/150` — i.e. 120 of 150 stops genuinely had a sibling parked on a marker instruction, and every one of the 150 marker calls was caught.

The bug also reproduced 4/4 and 3/4 on earlier runs ([31324476072](https://github.com/bingosuite/bingo/actions/runs/31324476072), [31323980359](https://github.com/bingosuite/bingo/actions/runs/31323980359)), including **without `-race`**.

## The test

`overlap`-labelled, linux-only, wired into the existing `Linux amd64 (ptrace)` job. One pinned thread hammers `MARK_A` while N pinned siblings hammer `MARK_B` (so siblings also cover the same-address case). After **every** stop it asserts both breakpoints are still installed; `breakpointTable.set` consults `byAddr` before touching tracee memory, so a successful `SetBreakpoint` cannot be an artifact of the probe, and a detection is corroborated read-only via `ClearBreakpoint` on the original id.

Non-vacuity is enforced, not assumed: at every stop it reads `/proc/<pid>/task/*/status` plus each thread's stopped PC from `/proc/<pid>/task/<tid>/syscall` and counts threads parked **on** a marker instruction — addresses resolved from the target's own DWARF, matched at `addr` or `addr+1` since an amd64 `INT3` stop leaves RIP one byte past the trap. A run that never reached two is rejected.

## Notes for review

- The pre-existing spurious-SIGTRAP pass is only PC-neutral **on amd64**, because `archRewindPC` (−1) and `len(archTrapInstruction())` (+1) cancel. A future linux/arm64 backend would advance by 4 and skip an instruction. Flagged in the commit message and worth its own issue if that port ever happens.
- **Out of scope, deliberately:** while a trap is lifted for a step-over, a thread executing that same address can still miss the breakpoint. That is inherent to software breakpoints without a stop-the-world and predates this change, so the spec counts hits for non-vacuity but never requires a particular total.
- Two independent adversarial reviews confirmed the causal chain and the detector. They split on the remedy — one accepted the engine-side rearm alone, the other blocked it as incomplete because it abandons an in-flight `PTRACE_SINGLESTEP`. This PR takes the union: fix the root cause in the backend *and* keep the engine rearm for backends that can still deliver such a stop.
- `AGENTS.md` is updated in the same commit (step-over atomicity invariant, the linux targeted wait, and the new label), per the repo's drift rule.
- `scratch-198-unfixed` is the A/B control branch and can be deleted after review.

### DIFF
--- a/.github/workflows/debugger-e2e.yml
+++ b/.github/workflows/debugger-e2e.yml
@@ -125,6 +125,12 @@ jobs:
         if: always()
         run: go test -tags e2e -race -count=1 -v -timeout 900s ./test/integration -ginkgo.v -ginkgo.label-filter=overlap
 
+      # Same spec WITHOUT the race detector: removes the "-race changed the
+      # scheduling" objection to the reproduction.
+      - name: E2E overlap run 4 (no -race)
+        if: always()
+        run: go test -tags e2e -count=1 -v -timeout 900s ./test/integration -ginkgo.v -ginkgo.label-filter=overlap
+
   darwin-arm64:
     name: Darwin arm64 (Mach exceptions)
     runs-on: macos-14
--- a/.github/workflows/debugger-e2e.yml
+++ b/.github/workflows/debugger-e2e.yml
@@ -87,6 +87,44 @@ jobs:
       - name: E2E concurrency (goroutine/thread snapshot foundation)
         run: go test -tags e2e -race -count=1 -v -timeout 300s ./test/integration -ginkgo.label-filter=concurrency
 
+  # TEMPORARY PROOF JOB — do not merge. workflow_dispatch only, so PR/push
+  # semantics of every other job are untouched. Drives the `overlap` label: a
+  # sibling thread's software-breakpoint stop delivered by Wait4(-1, WALL) while
+  # the engine is mid-step-over of another thread's breakpoint.
+  overlap-proof-linux-amd64:
+    name: PROOF overlap (sibling BP during step-over, ptrace)
+    if: github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 40
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.25.5"
+
+      - name: Confirm native target
+        run: test "$(go env GOOS)/$(go env GOARCH)" = "linux/amd64"
+
+      - name: Environment
+        run: |
+          uname -a
+          nproc
+          cat /proc/sys/kernel/yama/ptrace_scope || true
+          go version
+
+      - name: E2E overlap run 1
+        run: go test -tags e2e -race -count=1 -v -timeout 900s ./test/integration -ginkgo.v -ginkgo.label-filter=overlap
+
+      - name: E2E overlap run 2
+        if: always()
+        run: go test -tags e2e -race -count=1 -v -timeout 900s ./test/integration -ginkgo.v -ginkgo.label-filter=overlap
+
+      - name: E2E overlap run 3
+        if: always()
+        run: go test -tags e2e -race -count=1 -v -timeout 900s ./test/integration -ginkgo.v -ginkgo.label-filter=overlap
+
   darwin-arm64:
     name: Darwin arm64 (Mach exceptions)
     runs-on: macos-14
--- a/.github/workflows/debugger-e2e.yml
+++ b/.github/workflows/debugger-e2e.yml
@@ -9,7 +9,9 @@ name: Debugger E2E
 # suite lives in ./test/integration and is split via Ginkgo labels into `basic`
 # (correctness), `churn` (multi-thread robustness), `pause` (async interrupt /
 # manual stop), `stepping` (StepInto/StepOut), `inspect`
-# (StackFrames/Locals/Goroutines), `breakpoints` (ClearBreakpoint), `kill`
+# (StackFrames/Locals/Goroutines), `breakpoints` (ClearBreakpoint), `overlap`
+# (a sibling thread's breakpoint stop landing during another thread's
+# step-over — linux-only, see #198), `kill`
 # (kill-while-running), `exit` (process exit status), `restart`, `fullstack`
 # (operations driven through the whole client → WebSocket → hub → debugger stack),
 # and `dap` (a real Debug Adapter Protocol client driving a session over TCP,
@@ -75,6 +77,9 @@ jobs:
       - name: E2E breakpoints (ClearBreakpoint)
         run: go test -tags e2e -race -count=1 -v -timeout 300s ./test/integration -ginkgo.label-filter=breakpoints
 
+      - name: E2E overlap (sibling breakpoint during a step-over)
+        run: go test -tags e2e -race -count=1 -v -timeout 600s ./test/integration -ginkgo.label-filter=overlap
+
       - name: E2E kill (kill-while-running)
         run: go test -tags e2e -race -count=1 -v -timeout 300s ./test/integration -ginkgo.label-filter=kill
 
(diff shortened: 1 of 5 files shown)


xsachax:
Closing this draft because adversarial state-machine review found the targeted `Wait4(stepTID)` remedy unsafe: when the stepped TID is the thread-group leader it reintroduces the delayed-leader `Wait4(pid)` reaping hazard fixed in #111, and absorbed exact-TID ptrace events can cancel the hardware step while leaving `stepping` latched. The engine-side fallback can also abandon an active step. The deterministic reproduction remains valid; the replacement backend-local Wait-owned FIFO is tracked in #199.

xsachax:
Superseded by #199 / #202, which are canonical for this path.

The native-proof evidence, full repeat-rate table, minimal reproduction, and two corrections to the write-up have been added to #199 and #202 as comments.

**These refs are supporting evidence and must not be deleted:**

- `xsachax-proof-linux-sibling-bp-overlap` @ `fafe0a0` + tag `issue-198-proof-evidence` — original native proof (reproduced 10/12 attempts, incl. without `-race`)
- `scratch-198-unfixed` @ `13f1e5a` — A/B control: the same spec on unfixed `main`, fails in 2.7s
- `scratch-198-minrepro-unfixed` @ `99e12dd` / `scratch-198-minrepro-fixed` @ `9edf85a` — 182-line, 2-thread minimal repro (2/6 unfixed, 6/6 fixed)
- `fix-linux-sibling-bp-step-over` @ `5323852` — the alternative fix that ran green on all 13 checks; kept for comparison only

Run IDs: [31323712681](https://github.com/bingosuite/bingo/actions/runs/31323712681), [31323980359](https://github.com/bingosuite/bingo/actions/runs/31323980359), [31324476072](https://github.com/bingosuite/bingo/actions/runs/31324476072), [31325537802](https://github.com/bingosuite/bingo/actions/runs/31325537802) (fixed, green), [31325539248](https://github.com/bingosuite/bingo/actions/runs/31325539248) (unfixed control), [31326417720](https://github.com/bingosuite/bingo/actions/runs/31326417720) / [31326418749](https://github.com/bingosuite/bingo/actions/runs/31326418749) (minimal repro A/B).

Issue #94, opened 2026-07-19
Linux backend reports ExitCode 0 for every tracee — real exit status dropped at PTRACE_EVENT_EXIT


xsachax:
**Describe the bug**

On the linux/amd64 ptrace backend, `EventProcessExited` always carries `ExitCode: 0`, regardless of the tracee's real exit status. A traced Go program that exits with a non-zero code (e.g. `os.Exit(3)`) is reported to clients as a clean exit-0, and a process that dies from a fatal signal is likewise mislabeled as a normal exit-0.

The darwin/arm64 backend reports the true code (`StopEvent{... ExitCode: ws.ExitStatus()}`, `backend_darwin_arm64.go:888`), so this is a genuine linux-only defect and a cross-platform inconsistency. Nothing in AGENTS.md documents linux exit-code-0 as intentional.

**To Reproduce**
1. Build a trivial target: `package main; import "os"; func main() { os.Exit(42) }`.
2. `Launch` it under the debugger (linux/amd64), wait for the initial stop, then `Continue`.
3. Observe the `EventProcessExited` payload.
4. `ExitCode` is `0`, not `42`.

**Expected behavior**

`EventProcessExited.ExitCode` should be the tracee's real exit code (`42`), matching the darwin backend and the wire contract (`protocol.ProcessExitedPayload.ExitCode`). A signal-terminated tracee should surface as killed, not as a clean exit-0.

**Root cause**

`internal/debugger/backend_linux_amd64.go`, `Wait()`, the `PTRACE_EVENT_EXIT` case for the main thread hardcodes the exit code:

```go
// Main thread is about to call exit_group — let it actually exit.
if err := b.continueIfTraceeExists(tid, 0); err != nil { ... }
b.recordStop(tid)
return StopEvent{Reason: StopExited, TID: tid, ExitCode: 0}, nil
```

`startTracedProcess` sets `PTRACE_O_TRACEEXIT`, so the kernel stops the main thread at `PTRACE_EVENT_EXIT` **before** it actually dies. The engine tears down on this `StopExited`, so the real status never surfaces via a later `wait4` `Exited()`/`Signaled()`. Because `PTRACE_O_TRACEEXIT` is inherited by every clone, the group leader deterministically hits this branch first, so the reported code is effectively always `0`.

[...]

PR #202, opened 2026-08-09
fix(debugger): park foreign-thread stops during a linux single-step

### LINKED ISSUE #199: [debugger] Linux sibling breakpoints corrupt an in-flight single-step

xsachax:
Fixes #199

> **Stacked PR.** The base is `xsachax-fix-async-engine-suspension-halt` (#184), not `main`.

## Problem

Linux waits with `Wait4(-1, ..., WALL)`, so a sibling thread can stop while
another thread is in the software-breakpoint transaction:

```
restore original bytes -> single-step owner -> reinstall trap
```

Returning that sibling stop to the engine before the single-step completes
breaks the engine/backend current-stop pairing and can lose the original
breakpoint while its trap is disarmed. A same-address sibling can instead be
misclassified as an unowned trap and resume in the middle of the restored
instruction.

Two related races must also be closed:

1. The stepped thread can exit while sibling stops are held. Clearing the
   backend step bit alone opens the queue while the engine still owns the
   removed breakpoint.
2. A delayed hit on an auto-cleared one-shot sentinel can arrive after the
   sentinel is restored. Address history alone is not proof that this is that
   stale hit: genuine `INT3`/`ICEBP`/`INT 3` or `BRK` instructions can later
   occupy the same address.

### Park foreign stops in `Wait`

`linuxBackend.Wait` parks only user-visible `StopBreakpoint` and `StopSignal`
events that belong to a thread other than the active `stepTID`. The thread
remains ptrace-stopped and the whole `StopEvent` is held FIFO.

- The stepped thread's own stop always reaches the engine so the trap can be
  reinstalled.
- `recordStop` runs only when an event is delivered, never when it is parked.
  TID-less memory and resume operations therefore still target the current
  engine stop.
- The queue drains only when no step or dead-owner reconciliation is pending.
- Inline clone/group-stop/preemption handling remains inline. If an absorbed
  stop consumed the owner thread's step, it is re-armed with
  `PTRACE_SINGLESTEP`; exec and unknown ptrace events abort rather than run with
  an un-restored breakpoint.
- Main-process teardown purges the queue inside `Wait`. There is no
  engine-callable queue purge and no queue mutex: one-shot `Wait` calls are
  serialized by the engine handoff.

### Reconcile a dead step owner before releasing siblings

Thread death closes hardware-step ownership but keeps the park gate closed.
`Wait` then returns one internal `StopStepThreadExited` boundary anchored to the
oldest held TID without dequeuing its real stop. If no sibling is held yet, the
first later user-visible stop becomes that anchor.

The engine handles the boundary before normal stop processing:

1. reinstall `steppingOverBP` through the current stopped TID;
2. acknowledge the boundary to the backend;
3. start the next one-shot `Wait`, which may now deliver the FIFO head.

If reinstall fails, the engine emits `Error -> Paused`, retains the transaction
and keeps the gate closed. `Kill`/`Restart` remains the safe recovery; a sibling
cannot silently take ownership first.

As a cross-platform invariant, the ordinary `StopBreakpoint` branch also
reinstalls any still-lifted `steppingOverBP` before breakpoint lookup. Linux
normally reaches that branch only after the boundary has already reconciled it,
so this is a no-op there. It is the backstop for Darwin, whose Mach backend
re-faults foreign traps but has no Linux park queue or boundary event. The
same-address case is resolved only after the trap is back in the table.

### Recover delayed sentinels only with byte provenance

When an internal one-shot breakpoint is successfully restored, the engine
records the restored instruction bytes for that runtime address. An unowned
trap at a retired address is recovered only when:

1. current memory is not a live architecture trap spanning the reported stop;
2. current bytes exactly match a byte sequence this engine restored there.

The architecture decoders cover:

- amd64: `CC`, `F1`, `CD 03`, including the cross-boundary `CD 03` stop;
- arm64: every `BRK #imm16` encoding.

A genuine live trap is surfaced through the normal unowned-trap path instead of
being rewound forever. A changed non-trap byte also does not qualify as stale.
For a proven delayed sentinel, the engine rewinds the live registers on the
exact stopping TID and resumes it. Read/register/resume failures use the
suspending `Error -> Paused` path.

This PR does not take #221's pending-signal/SIGURG state.

### Unit and mutation gates

- Host-agnostic classifier, FIFO, death-gate, boundary, resume-plan and abort
  tests in `waitpark_test.go`.
- Scripted production `Wait` branches for `PTRACE_EVENT_EXIT`, reaped thread
  death, and the "no held stop yet" ordering.
- Exact signal-value gates cover both production copies: park-queue capture and
  direct delivery with no step or the owner thread's own signal.
- Exact current-stop/TID tests prove the reconciliation write and delayed
  sentinel `SetRegisters` mutation target the delivered sibling.
- Engine ownership tests prove the original breakpoint remains installed and
  clearable after the owner dies.
- Genuine-trap tests cover all supported amd64/arm64 encodings plus changed
  restored bytes. Forcing the restored-byte predicate true makes the focused
  changed-byte test fail.
- Reinstall/read/register/resume fault injection proves asynchronous failures
  produce `Error -> Paused` rather than stranding the hub.

### Native Linux gates

The `overlap` label contains seven specs:

- source step-over overlap;
- machine-step overlap;
- foreign `SIGUSR1` overlap with signal-value preservation;
- `Pause` interrupt overlap;
- breakpoint-free `Pause` delivery backstop;
- kill with held stops;
- raw `SYS_exit` of the exact step owner while a sibling stop is parked.

The specs assert non-vacuity through cumulative backend counters. The plain
overlap requires both `LinuxParkedStopCount > 0` and
`LinuxRetiredInternalBreakpointCount > 0`; the owner-death spec requires both
`LinuxStepThreadExitCount > 0` and parked stops. Churn runs five independent
rounds.

Exact-head native run
[31359224572](https://github.com/bingosuite/bingo/actions/runs/31359224572),
Linux job
[93364611244](https://github.com/bingosuite/bingo/actions/runs/31359224572/job/93364611244):

- all Linux labels passed, including five churn rounds;
- overlap passed 7/7 specs;
- plain overlap: 90 cycles, 36 parked stops, 14 recovered delayed sentinels;
- machine-step overlap: 28 parked stops;
- signal overlap: 23 parked stops, 3 parked signal stops, exact signal map
  `map[10:7]`, 4 native step re-arms;
- pause overlap: 54 accepted pauses, 29 surfaced pause events, 4 interrupts
  held mid-step, 21 parked stops / 9 parked signal stops;
- dead-owner proof: 1 exact step-owner exit and 1 parked sibling stop.

## Review

Two independent adversarial review passes were run against the production
design:

- dead-owner boundary / exact-current-stop ownership;
- retired-sentinel provenance / architecture trap decoding.

Both final-head reviews found no significant issues. They verified the ordered
Linux boundary, cross-platform `StopBreakpoint` backstop, byte-provenance
discriminator, architecture decoders, exact-TID mutations, Error -> Paused
paths, direct-delivery signal gate, three-phase handoff assertion, and native
non-vacuity.

### Fresh max-rigor adversarial review at `8a33fb3` — **no blockers**

Head-verified first (`git rev-parse HEAD` = `8a33fb3`), reviewed `git diff base...HEAD`, and explicitly did **not** reason about the rival branch's SHAs. It ran the linux container suite and the darwin native suite, and applied **its own** production mutation (`resumeFor` → always `resumeContinue`, `diff -q`-confirmed applied): the linux suite failed, then restored clean. The freeze-prevention rule is genuinely discriminated, by an independent mutation rather than only mine.

Traced and confirmed:

- **All 9 inline resume sites** (clone, non-main `EVENT_EXIT`, main `EVENT_EXIT`, exec, user SIGTRAP, new-thread SIGSTOP, SIGURG, SIGCONT, other-signal) route through `absorbStop`/`planAbsorb`/`applyAbsorb`, so a stop on `stepTID` is re-armed, resolved, or fails safe. **No site leaves `stepping == true` with the step silently cancelled** — the blocker-1 claim, checked directly against this branch's code.
- The lone bare `continueIfTraceeExists` **is** the terminal main-thread `PTRACE_EVENT_EXIT`; it purges and returns `StopExited`, so a latched step is irrelevant.
- Every process-exit/kill path returns `StopExited`/`StopKilled` **regardless of `stepExitPending`**; no teardown resumes a parked or dead TID.
- **G4 holds**: `recordStop` never runs in `park()`, so `traceTID()` always names a genuinely ptrace-stopped thread and a parked TID can never become the `ContinueProcess` target.
- **The engine reconcile is single-reinstall and correctly ordered.** On linux a parked breakpoint drains only after `steppingOverBP` is already nil, so it is a no-op there and active only on darwin's re-fault path — no double-reinstall. Reinstall failure correctly does **not** acknowledge the boundary, emits `Error → Paused`, and rejects Continue/Step; Kill still works because the engine loop services `cmdCh` while `Wait` blocks. A documented recoverable fail-safe, not a hang.

**Its one MINOR** (explicitly *not* filed as a bug): `stepExitPending` stays set if the stepped TID dies with an **empty** queue, since the boundary needs a held stop to anchor to. Documented in `cb9892f` rather than changed, because the constraint is forced by ptrace: the boundary exists to give the engine a TID it can `POKEDATA` into, and ptrace can only write through a thread that is actually stopped — an empty queue offers no legal anchor. The flag simply persists until the first foreign stop parks and supplies one; whole-process death tears down normally; and the step-over case cannot reach it at all, because the stepped instruction is user code at `bp.addr` and a thread cannot exit executing it. `AGENTS.md` rule 5 now records this so a change making *plain* foreign single-steps common re-checks it.

### Native darwin verification

`just e2e-darwin` on real Apple Silicon: **22/22 specs, all labels**, including `hygiene` (task send refs `base:1 / after:1` — no Mach port leak). `darwin-e2e-verified` applied; the gate re-ran green ([31359494591](https://github.com/bingosuite/bingo/actions/runs/31359494591)).

### One darwin `churn` flake, investigated and attributed away from this PR

The first `just e2e-darwin` at `cb9892f` failed 1/22: `churn`, at `debugger_e2e_common_test.go:426` — the `Continue` → `Expect(BreakpointHit)` assertion. Reported at **426 (the `Expect`)** rather than **424 (the `waitFor`)**, so it was a genuine wrong-kind event, not a harness timeout. On darwin a crashing tracee surfaces as a **process exit, not a signal stop**, so that assertion is exactly where a byte-patching bug would appear — it could not be waved off.

**It did not reproduce**, and the paths it would have to travel are provably not taken:

| Evidence | Result |
| --- | --- |
| Isolated `churn` runs after the failure | **33/33 green** |
| Amplified `churn` (`BINGO_E2E_CHURN_ITERS=1000`) | **8/8 green — 8,000 iterations** |
| Full 22-spec darwin suite, HEAD | **7/7 green** |
| Full 22-spec darwin suite, base `45dfa4f` | **4/4 green** |
| Total post-failure churn iterations | **~16,000, zero repro** |

Non-reproduction alone proves nothing, so the reachability was measured directly. Both production paths this PR adds that are *live on darwin* were instrumented and counted during churn:

| Instrumented path | Hits |
| --- | --- |
| `INSTRUMENT_POSITIVE_CONTROL` (top of `case StopBreakpoint`) | **400** |
| `handleStop` step-over reconcile (this PR) | **0** |
| `resumeRetiredInternalBreakpoint` / `archLiveTrapResumePC` | **0** |

The positive control confirms the instrument really reports (400 hits = 200 iterations × 2 stops); in that same run both new paths were **never entered**. Consistent with the static picture: the churn target is **byte-identical to base**, its 180s watchdog line predates this PR, and the park queue is referenced **zero** times by `backend_darwin_arm64.go` (`waitpark.go` is dead code on darwin, compiled there only to keep it mutation-checkable from macOS).

Conclusion: **not a regression from this PR.** The same symptom — an unexpected `ProcessExited` in an *unmodified* `churn` — is already recorded in the issue evidence at run [31323980359](https://github.com/bingosuite/bingo/actions/runs/31323980359). Flagging it as a **pre-existing rare darwin flake** worth its own issue rather than silently absorbing it here; I did not widen this PR to chase it.

## Validation

Frozen head: `8a33fb37eee2ebc2bf2a99b137ff79b7e6df4c3d`

| Gate | Result |
|---|---|
| `go test -tags bingonative -race -count=1 ./...` | pass |
| `go vet -tags bingonative ./...` | pass |
| Linux E2E-tagged vet + linux/amd64 debugger/E2E cross-builds | pass |
| Native Darwin full suite, CI-sized churn (`BINGO_E2E_CHURN_ITERS=120 just e2e-darwin`) | 22/22 pass; churn 120/120 |
| Go workflow [31359224562](https://github.com/bingosuite/bingo/actions/runs/31359224562) | build + unit tests pass |
| VS Code workflow [31359224560](https://github.com/bingosuite/bingo/actions/runs/31359224560) | linux-x64 + darwin-arm64 pass |
| Debugger E2E [31359224572](https://github.com/bingosuite/bingo/actions/runs/31359224572) | all jobs pass; native Linux overlap 7/7 and churn 5/5 |
| Darwin verification gate [31359494591](https://github.com/bingosuite/bingo/actions/runs/31359494591) | pass after exact-head local proof |

The unbounded local Darwin default uses 200 churn cycles against a target with
a fixed 180-second watchdog and crossed that watchdog at iterations 179 and
171. The clean run uses the repository's documented/CI-sized 120 cycles, which
completed in 101 seconds with the required headroom.

---

## Follow-up hardening (final head `5db67f6cc28ebb8c1f3dbc559f315070b470d873`)

One additional commit, `fix(debugger): anchor step-exit reconciliation and fail closed on exec`, closes two gaps found reviewing the reconciliation boundary above.

**1. The empty-anchor case was reachable, and the doc rationale for it was false.**
The boundary needs a genuinely ptrace-stopped TID to `POKEDATA` the lifted trap back through. With a sibling parked it borrowed that sibling's TID; with **nothing** parked there was no anchor at all, so the owner was resumed and the reinstall waited for whatever stop happened to arrive next — unbounded, and on a quiet tracee never. AGENTS.md argued the case was unreachable because a thread cannot exit executing the instruction at `bp.addr`. A breakpoint can sit directly on a raw `SYS_exit`, so it can.

`Wait` now **holds** the dying owner at its `PTRACE_EVENT_EXIT` stop instead of resuming it (that stop fires in `do_exit` before `exit_mm`, so its address space is still mapped) and anchors the boundary on it. It is released exactly once, by `completeStepThreadExit`, and only **after** the engine has reinstalled. A parked sibling is still only ever borrowed, never resumed. Because the release is a real ptrace op, `completeStepThreadExit` now returns an error: `ESRCH`/`ENOENT` is benign, any other failure keeps both the hold and the gate so the engine halts suspended (`Error → Paused`, no new `waitLoop`) and `Kill`/`Restart` stays the recovery.

**2. `PTRACE_EVENT_EXEC` is now unconditionally fatal.**
It was fatal only on the stepped thread and silently `PTRACE_CONT`ed otherwise. `execve` replaces the image for **every** thread, so each breakpoint address, saved instruction byte and tracked tid describes memory that is gone — there is nothing to reconcile against regardless of which thread reports it or whether a step was in flight. It is no longer an absorb decision at all (`absorbExec` is removed; `absorbUnknownEvent` stays TID-keyed), and the error no longer implies the reported pid is the stepped tid — after `execve` the kernel reports the thread-group leader's. bingo's own startup stays out of reach because the launch path consumes its `execve` stop before enabling `TRACEEXEC` and attach enables no options.

Known consequence, documented: a tracee launched through a **binfmt interpreter that re-execs** (Rosetta under emulated linux/amd64 Docker; some qemu-user setups) now fails fast. Verified with a standalone ptrace probe — the exec'ing tid's `/proc/<tid>/exe` was `/run/rosetta/rosetta`. The old "success" was coincidence: the entry stop and DWARF slide describe the pre-exec image either way.

### Test-gate correction

`declareStepOwnerExitSpec` asserted `parked > 0` as its non-vacuity gate. With the owner-held path that transaction now resolves in microseconds, so on a fast runner **no sibling ever needs to park** — CI run [31386588498](https://github.com/bingosuite/bingo/actions/runs/31386588498) reported `parked-stops 0, held-owners 1`. Both anchors are correct, so the gate is now `parked + held > 0` and each count is reported. The new `declareStepOwnerHoldSpec` pins the owner-held path exclusively, with no sibling breakpoint armed at all.

### Validation at `5db67f6`

| Gate | Result |
|---|---|
| `go test -tags bingonative -race -count=1 ./...` (macOS) | pass |
| `go vet -tags bingonative ./...` / `gofmt` / `goimports` | clean |
| linux/amd64 `go vet ./...`, `go vet -tags e2e ./test/integration`, `go test -race` | pass |
| `golangci-lint run` (v2.12.2, CI version) | 0 issues |
| Deliberate mutants (11: unconditional exec, hold omission, early release, parked-sibling release, ignored release error, skipped reinstall, purge keeps hold, TID-keyed unknown-event, hold-but-resume, boundary ignores held owner, release/gate order) | all killed, each with a named killer |
| Native `just e2e-darwin` (exact head content, quiet host, single run) | 22/22 pass in 252s |
| Go [31388186313](https://github.com/bingosuite/bingo/actions/runs/31388186313) · VS Code [31388186320](https://github.com/bingosuite/bingo/actions/runs/31388186320) | pass |
| Debugger E2E [31388186316](https://github.com/bingosuite/bingo/actions/runs/31388186316) | all 4 jobs pass |
| Darwin verification gate [31389639050](https://github.com/bingosuite/bingo/actions/runs/31389639050) | pass |

Native non-vacuity from run 31388186316: `overlap-parked-stops 28`, `stepinto-parked-stops 20`, `signal-parked-stops 17` / `parked-signal-stops 3` / `signal-values map[10:14]` (SIGUSR1) / `step-rearms 4`, `step-owner-exits 1` with `parked 0 + held 1`, and `held-step-owners 1` / `held-step-owner-exits 1` for the new spec.

### Deliberately out of scope

- Leader `PTRACE_EVENT_EXIT` is still inferred as whole-process death (a leader can raw-`SYS_exit` while siblings survive). Wholly pre-existing; belongs with the #205 wait-ownership work.
- No `Wait`-error path kills or retains the tracee, so `closeTracer` detaches and the tracee runs on untraced. Pre-existing for every `Wait` error; rule 9 makes it reachable more often. Recorded in AGENTS.md.
- Attach still installs no ptrace options, so an attached process's `exec` arrives as a plain `SIGTRAP`. Pre-existing; out of scope per #221/#225 boundaries.

---

## Scope correction: two further items (final head `7df3ebca...`)

A second commit, `fix(debugger): stop trusting a leader-id exit and discharge fatal stops`, closes two more holes in the same state machine. All four items are one contract.

**A leader-id `PTRACE_EVENT_EXIT` is not proof of death.** `de_thread()` retires the old leader when a **non-leader** calls `execve()`: the execing thread takes over the leader's pid and the retired task's exit arrives under `tid == pid` while the process runs on. A leader that merely `pthread_exit()`s is the same shape. Measured on Linux 6.10 the retired leader's `GETEVENTMSG` decodes to an ordinary *"exited, code 0"* — indistinguishable from a real exit — and the genuine death arrives much later, after the exec:

```
(LEADER-ID) EVENT_EXIT  msg -> exited code 0     <- process ALIVE
siblings    EVENT_EXIT + real exits
(LEADER-ID) EVENT_EXEC  msg = execing thread's FORMER tid
(LEADER-ID) EVENT_EXIT + real exit, then ECHILD  <- the actual death
```

`Wait` now stashes the status, resumes through the ordinary dying-thread absorb — the leader can be the thread under a single step, and a bare continue there latches the step gate and freezes the wait loop — and commits only on real group death or `ECHILD`, retracting on exec.

**An observed `wait4` status beats the stash.** `GETEVENTMSG` is that *task's* `do_exit` value; the leader's wait status carries the *group* exit code. A `main` that `pthread_exit(0)`s while a worker later `exit(3)`s or `abort()`s is reaped as 3 / signal 6, so preferring the stash would report a clean `0` — the #94 misreport in a new disguise. The stash is the `ECHILD` fallback, which is why the read must still happen before the leader is resumed.

**Fatal branches were abandoning the stops they own.** `abortStep` drops the queue's *record* of parked threads but leaves them ptrace-stopped, as it does the thread whose stop failed; `tracerThread.close()` then makes the kernel detach and **resume** them straight into breakpoints still in their text. Both branches now report `ErrSessionInvalidated` and the engine discharges the tracee explicitly. The exec case skips the byte restore via `ErrImageReplaced` — those bytes describe an image that no longer exists and would corrupt the new one.

### Corrections made during review

- An earlier AGENTS.md claim of mine — *"a parked thread cannot die underneath the queue"* — was **false**, and my supporting probe was inconclusive (it never printed stop events). `TASK_TRACED` carries `TASK_WAKEKILL`: a parked worker, untouched by the tracer, moved on its own to `EVENT_EXIT` the instant the group was `SIGKILL`ed. Both consumers of a parked TID already tolerate `ESRCH`, so no code changed; the invariant text now states the measured behaviour and forbids adding a liveness assumption.
- The exit-code precedence was inverted from my original design, per the counterexample above.

### Validation at `7df3ebc`

| Gate | Result |
|---|---|
| macOS `go test -tags bingonative -race ./...`, vet, gofmt/goimports | pass / clean |
| linux/amd64 vet (incl. `-tags e2e`), full `-race` suite, repeated focused runs | pass |
| `golangci-lint run` (v2.12.2, CI version) | 0 issues |
| Deliberate mutants | **22, all killed**, each with a named killer (M21 survived first pass and a killer was added) |
| Native `just e2e-darwin` (this exact tree, quiet host, single run) | 22/22 in 251.4s |
| Debugger E2E [31395672974](https://github.com/bingosuite/bingo/actions/runs/31395672974) | all 4 jobs pass |
| Go · VS Code · Darwin verification gate | pass |

Native non-vacuity for the new rule, from the CI log:

```
overlap-leader-retirement-message:
debugger: session invalidated: process image replaced (PTRACE_EVENT_EXEC reported
under pid 9045, execing thread's former tid 9049, retiring the previous leader): ...
```

`retiring the previous leader` only appears when a leader exit was actually stashed **and then retracted** — i.e. the real `de_thread()` sequence occurred on the CI kernel and the deferral fired. Pre-fix that run reported `ProcessExited(0)` for a live process.

### Still out of scope, documented

- Attached-tracee release is explicitly **not** claimed clean: a thread parked at a software breakpoint has its PC one trap-width past the trap on amd64 and `abortStep` has dropped the metadata to rewind it, and `bps.clearAll` ignores individual restore failures. A launched tracee is killed so it cannot matter.
- Ordinary `wait4` failures still detach-and-resume; only the two marked branches discharge (#205/#217).
- `EventProcessExited` for ordinary exits is now emitted at real group death rather than at the leader's `EVENT_EXIT`. The exit code is unchanged.


xsachax:
## ✅ Validation results

### Native linux/amd64 CI — [run 31326470997](https://github.com/bingosuite/bingo/actions/runs/31326470997), job `Linux amd64 (ptrace)` (7m18s)

Every label green, including the new one:

`basic` · `churn` · `pause` · `stepping` · `inspect` · `breakpoints` · `kill` · `exit` · `attach` · `concurrency` · **`overlap`** ✅

The `overlap` spec is strongly non-vacuous — it took 155s of its 600s budget and its report entries show the deferral path is genuinely being driven, not skated past:

| Metric | Value |
|---|---|
| `overlap-iterations` | 150 |
| `overlap-hits-A` | 106 |
| `overlap-hits-B` | 125 |
| `overlap-goroutines` | 5 |
| **`overlap-stepover-resolved-as-breakpoint`** | **82** |

That last one is the key number: on **82 of 150 cycles** a `StepOver` completed by delivering a *deferred sibling breakpoint* instead of its own `EventStepped` — i.e. the exact #199 interleaving occurred on more than half the iterations, and every time both logical breakpoints stayed armed, every hit resolved to a known id, and both ids were still clearable by their original ids at the end. Under the bug this is precisely where the entry is lost and `ClearBreakpoint` fails.

### Other CI on this PR

| Check | Result |
|---|---|
| Go (build, vet, unit, `-race`) | ✅ success |
| VS Code extension | ✅ success |
| Darwin verification gate | ✅ success (no darwin-native paths touched) |
| Debugger E2E — darwin + fullstack + dap jobs | compile/codesign ✅; execution skipped on hosted runners as designed |

### Native darwin/arm64, run locally on Apple Silicon

`just e2e-darwin` (codesigned, all labels, `-race`): **22 of 22 specs passed in 258.8s**, including `basic`, `stepping`, `breakpoints`, `churn`, `kill`, `exit`, `attach`, `concurrency`, `pause`, `inspect`, `restart`, `fullstack`, `dap`, and the darwin-only `hygiene` (task send refs `{base:1, after:1}` — no leak).

This is the regression evidence for the platform CI cannot execute. Darwin never defers anything (its receive loop already parks stragglers), so this run confirms the new code is inert there rather than perturbing the Mach path.

### Local (macOS)

- `gofmt -l` / `goimports -l` clean across `internal`, `test`, `pkg`, `cmd`
- `go vet -tags bingonative ./...`, `GOOS=linux GOARCH=amd64 go vet ./...`, and `GOOS=linux GOARCH=amd64 go vet -tags e2e ./test/integration` all clean
- `go test -tags bingonative -race -count=1 ./...` — all packages green
- Focused deferral specs re-run 5× under `-race` — green every time
- Mutation checks A/B/C all fail the focused specs as intended (details in the PR body)

### Residual risk update

The "never executed on real Linux" caveat in the PR body is now **closed** — the `overlap` spec ran on a native hosted linux/amd64 runner, passed, and exercised the deferral path 82 times. It also fits comfortably inside the job timeout (155s of 600s), so no tuning of `BINGO_E2E_OVERLAP_ITERS` is needed.

The remaining documented risk is unchanged: the bounded multi-deferral degradation where a non-suspending deferred handler plus a still-queued stop can leave `traceTID` naming a running thread. It surfaces as a `haltOnError` suspending `EventPaused` (session stays controllable, no strand), and eliminating it entirely would require a TID-explicit `ContinueProcess` in the `Backend` interface — a backend change, deliberately out of scope here.

xsachax:
## Follow-up commit `384376d` — a second, related engine defect found and fixed

A post-implementation design review (thanks @/creator-session for the challenge) questioned whether an engine-only fix can be correct at all, on the grounds that `recordStop(foreignTID)` is overwritten by the exact step completion and that `ContinueProcess` has no TID parameter. Both premises are factually true, and both are already handled by this PR's design — the stepped thread is deliberately **held** ptrace-stopped so `traceTID` still names it, and the owed resume is discharged by `releaseParked` at exactly the point where the TID-less continue still points at that thread.

But chasing the challenge did surface a **genuine second defect**, in the same mechanism, which the original commit did not cover.

### The defect

`resumeTracee` re-points `traceTID` at the current thread with a marker `SingleStep` (`internalStepTID`, swallowed without a user event). That step can be issued on a thread **sitting on an armed software breakpoint**.

`ContinueProcess` onto a trap is safe — it reports an ordinary `StopBreakpoint`. `SingleStep` is not: it *executes* the trap, and `backend_linux_amd64.go`'s `cause==0` SIGTRAP classifier keys purely on `stepping && tid == stepTID`, so **a trap hit is indistinguishable from a step completion**. The `internalStepTID` branch swallowed the hit and continued from `bpaddr+1` — mid-instruction on amd64, and a lost user breakpoint.

Confirmed with a throwaway probe against the fake backend before fixing: `InternalStepTID == tidB` with a trap armed at tidB's PC, and after the trap-executing step **no event at all** was emitted, plus an extra continue.

**Reachability.** Requires an empty deferral queue, an owed parked-thread release, and a trap armed at the resumed thread's PC. The *breakpoint* resume path never takes it (it goes through `stepThreadOverBP`), which is why the 82 deferrals observed in the previous CI run were all safe. It is reachable via the deferred ordinary-signal auto-resume, deferred leftover-pause suppression, the spurious-trap resume, `Continue` with `lastBP == nil`, `stepOut`, and `sourceStepOver`.

### The fix (still engine-side; no backend edit)

Before spending the TID-less continue, `resumeTracee` calls `armedBreakpointAt(current)`. If a trap is armed there it disarms it and routes the resume through the **existing, already-proven step-over machinery** (`steppingOverBP` + `bpResumeContinue` + `stepThreadOverBP`). The existing `StopSingleStep` handler then reinstalls the trap and continues, emitting nothing — precisely the semantics an internal resume needs.

Three details are load-bearing:

- `internalStepTID` is deliberately **not** set on this path. The swallow branch runs *before* the reinstall block, so setting it would pre-empt the reinstall.
- The disarming `WriteMemory` must precede `releaseParked`, for the same reason as everywhere else in this PR: the write targets `traceTID`, which is still the parked thread. Releasing first aims it at a live thread → `ESRCH`.
- `internalStepTID` is now also cleared when a marker step ends by any non-`StopSingleStep` route, so a stale marker cannot swallow an unrelated later stop.

Every failure after the disarm rolls back through `rearmInternalStepOver`. A register-read failure degrades to "assume no trap" (logged) rather than erroring the stop.

### Alternatives considered and rejected

- **A linux `Wait`-side FIFO queue** (park foreign stops inside `Wait`, return them only after the exact step completes). Structurally the cleanest end state and it mirrors darwin's documented *"loop until the trap belongs to the stepping thread"* invariant — but it edits the most fragile file in the repo, is linux-only, invalidates this PR's CI evidence, and would force all the deterministic engine specs to become backend tests, which are not writable without a real tracee. Best as a **separate follow-up** on top of this PR, after which the engine deferral becomes a safety net rather than the primary mechanism.
- **Mirroring darwin by rewinding the foreign PC and continuing so it re-faults** — rejected outright: with the trap still armed the thread re-faults immediately and **spins** until the step completes.

### New regression coverage

- `resuming a deferred thread that is parked on an armed trap` — asserts the trap is stepped over rather than executed blind, the hit is never swallowed, and pins the exact backend op order via a new ordered op log on `fakeBackend`: `write:addrA` (reinstall) → `write:addrB` (disarm) → `continue` (release parked) → `step:tidB`.
- `a Pause interrupt surfacing on a foreign thread mid-step` — withheld while the step is in flight (no early reinstall, no phantom event), then still delivered as `EventPaused` with the trap rearmed, and the still-parked stepped thread properly discharged by the next `Continue`. Covers the `manualStopPending` semantics end to end.

Two further mutation checks, both confirmed load-bearing:
- disabling the armed-trap guard → the armed-trap spec fails;
- moving the disarm *after* `releaseParked` → the ordering assertion fails.

The deferral suite is now **10 deterministic specs**.

### Validation of `384376d`

- `gofmt`/`goimports` clean; `go vet -tags bingonative ./...` and `GOOS=linux GOARCH=amd64 go vet ./...` clean
- `go test -tags bingonative -race -count=1 ./...` — all packages green (macOS); `internal/debugger` repeated 5× under `-race`, stable
- `just e2e-darwin` (native Apple Silicon, all labels) — **22/22 SUCCESS**, 257s
- Native GitHub CI on this commit — **all four workflows green**: Debugger E2E (Linux ptrace, Darwin Mach, both full-stack legs), Go, Darwin verification gate, VS Code extension
- `overlap` label on this commit: 150 iterations / 156s, hits A=110 B=127 across 4 goroutines, **`overlap-stepover-resolved-as-breakpoint: 87`** — the deferral path fired on 87/150 cycles, so the spec is non-vacuous

`AGENTS.md` is extended in the same commit with the armed-trap invariant, the disarm-before-release ordering rationale, and the stale-marker clearing rule.

xsachax:
I produced the native proof for #199 in a parallel session. Two directly actionable items for this PR, plus independent evidence. No objection to the design — see the last section.

### 1. The new AGENTS.md prose enshrines an incorrect mechanism

> …takes the spurious-SIGTRAP path, **advances PC** and calls `ContinueProcess`…

On amd64 it does not advance PC. `populateBreakpointStop` → `populateStopPC(stop, true)` sets **`stop.PC = RIP − 1`** in a local; the live register is never touched. The spurious branch then writes `regs.PC = stop.PC + len(archTrapInstruction())` = `(RIP − 1) + 1` = **`RIP`** — a no-op write.

The actual defect is the **absence of `rewindToBreakpoint`** on that path (the matched path calls it; the spurious path does not), so the thread is resumed with RIP still at `addr+1`. Since this paragraph is becoming the canonical explanation, suggest:

> …finds no entry, takes the spurious-SIGTRAP path, which skips `rewindToBreakpoint` and calls `ContinueProcess` — resuming that thread at `addr+1`, one byte into the restored instruction — and which also clears `stepping`/`stepTID` so the real step completion is misclassified.

Two reviewers independently reached the same conclusion, and I confirmed it by disassembly. Worth adding the severity nuance too: whether that mid-instruction resume is harmful is **encoding-dependent**. In the proof target the line starts with a 7-byte `LEAQ`, and `addr+1` decodes as `lea edx` — RIP-relative and non-PIE, so it computes the *same* address and re-syncs, i.e. benign by coincidence. But a breakpoint on `atomic.AddInt64(d, c)` patches the `f0` **LOCK prefix** itself, and `addr+1` decodes as a plain, **non-atomic** `XADDQ`.

### 2. Two assertions a green `overlap` run will not catch

Learned the hard way while building the proof spec:

- **The rearm-vs-lookup ordering is invisible to an armed-ness probe.** Any spec whose invariant is "both breakpoints are still installed" passes whether the trap is restored before or after `bps.atAddr`. The assertion that actually distinguishes them is that **`SetRegisters` was called with the rewound PC** for a same-address sibling — i.e. it resolved as a real hit rather than via the spurious path. Cheap to pin with the existing `fakeBackend`.
- **The park queue's safety window is not the engine's danger window.** The backend clears `stepping` the instant the trap is reaped, but the engine stays disarmed until `bps.reinstall`. The gap is only safe because **`waitLoop` is one-shot and none runs while `loop()` is inside `handleStop`**. Turning `waitLoop` into a long-lived `for { Wait(); stopCh <- … }` goroutine — a natural performance refactor — would silently reopen #199 with no compile error and no failing test. A `fakeBackend` assertion that at most one `Wait` is in flight between the `WriteMemory(addr, original)` and the `WriteMemory(addr, trap)` would pin it.

Your *"exact per-cycle sibling hit counts are inherently racy — never assert them"* matches my experience exactly: my first spec required an exact per-cycle hit count and hung, because a thread can execute the address while the trap is lifted and legitimately not trap.

### 3. Independent CI evidence, and an A/B pattern you can reuse

Same-spec A/B on hosted ubuntu-24.04, real ptrace:

| Branch | Production code | `overlap` |
| --- | --- | --- |
| `scratch-198-unfixed` @ `13f1e5a` ([run](https://github.com/bingosuite/bingo/actions/runs/31325539248)) | main `0509b53` | **fails in 2.7s** |
| `scratch-198-minrepro-unfixed` @ `99e12dd` ([run](https://github.com/bingosuite/bingo/actions/runs/31326417720)) | main `0509b53` | **2/6 attempts fail** (182-line, 2-thread minimal repro) |

Building the *unfixed* control by plumbing (`git commit-tree` with the production files reverted to `main` and only the test carried over) makes "the spec really does fail without this change" a one-command check. Sibling count matters: 2 marker threads reproduce ~33% per attempt, 5 threads ~83% — worth ≥4 siblings in the permanent spec.

### 4. On the design — no objection

I evaluated an engine-side defer/queue and a per-TID state variant, and an adversarial review rejected both for reasons this PR has already designed around: engine-side replay cannot restore `lastStopTID` for `ContinueProcess` (your rule 4 — `recordStop` on delivery only), a dropped queue entry leaves a thread ptrace-stopped forever (your rules 5/6), and an engine-side queue covering only `StopBreakpoint` would leave a sibling `StopSignal` mid-step to tear the step-over down (you park both).

I also prototyped a backend variant that targets `wait4` at the stepping TID so sibling statuses are never dequeued at all. It is smaller, but **worse than this PR on liveness**: it stops absorbing clone/exec/`SIGURG` inline for the duration of the step, so a step over a blocking-syscall instruction stalls the whole tracee and defers `Pause` indefinitely. Your park queue keeps that absorption inline and only defers user-visible foreign stops, which is the better trade. Mentioning it only so the option is on record as considered and rejected.

xsachax:
Completing the record: my earlier comment linked only the **failing** half of each A/B. Here are both halves, so the comparison is verifiable end to end.

### Decisive A/B — same spec, same runner image, only the production code differs

| Spec | Production code | Branch @ SHA | Run | Result |
| --- | --- | --- | --- | --- |
| `overlap` (1×MARK_A + 4×MARK_B) | `main` `0509b53` | `scratch-198-unfixed` @ `13f1e5a` | [31325539248](https://github.com/bingosuite/bingo/actions/runs/31325539248) | **FAIL** in 2.7s — `SetBreakpoint(:25)` succeeded, `ClearBreakpoint(id=2)` → `breakpoint 2 not found` |
| `overlap` (identical spec) | path fixed | `fix-linux-sibling-bp-step-over` @ `5323852` | [31325537802](https://github.com/bingosuite/bingo/actions/runs/31325537802) | **PASS** — `stops=150 hitsA=30 hitsB=120 atMarkerPC(max)=4 overlapReady=120/150`; all 4 jobs green |
| `overlapmin` (2 threads, 182 lines) | `main` `0509b53` | `scratch-198-minrepro-unfixed` @ `99e12dd` | [31326417720](https://github.com/bingosuite/bingo/actions/runs/31326417720) | **2/6 attempts FAIL** (attempt 3 `-race`, attempt 5 no-`-race`) |
| `overlapmin` (identical spec) | path fixed | `scratch-198-minrepro-fixed` @ `9edf85a` | [31326418749](https://github.com/bingosuite/bingo/actions/runs/31326418749) | **6/6 attempts PASS** |

All four ran on hosted `ubuntu-24.04`, Go 1.25.5, kernel 6.17.0-1020-azure, real ptrace.

**Scope caveat, stated explicitly:** the two passing runs used the *alternative* fix from the now-closed #201, not this PR's park queue. So this A/B establishes that **the spec is a valid discriminator** — it fails on unmodified `main` and passes once the path is repaired — which is what qualifies it as a regression gate. It does **not** validate #202's implementation; that needs the same A/B re-run against this branch.

Also on the record, from the unfixed side: the non-vacuity instrumentation showed 4 threads simultaneously in ptrace tracing-stop with PCs at exactly `MARK_A+1`, `MARK_B+1`, `MARK_B+1`, `MARK_B` (addresses taken from the target's own DWARF; an amd64 `INT3` stop leaves RIP one byte past the trap), so the race really was set up rather than assumed.

xsachax:
The `lastStopTID` race found in run `31329090064` is now extracted as issue #231 / draft base PR #232 from `main`.

Please restack #202 on #232 and drop the incidental atomic commit `04f7f9a`; #202 should continue to own parked sibling-stop delivery and the rule that `recordStop` moves on delivery, while #232 owns only cross-goroutine field access.

Evidence for the base layer:
- unfixed production proof `31338374195`: stopped controls green; running Kill/Write/Read all race
- fixed `6ffa54b`: run `31338972574`, job `93309465646`, exact permanent selector green
- atomic-removal mutation `4f03d8c`: run `31339235577`, job `93310150239`, stopped control green and every running path caught
- all PR checks green; native debugger matrix `31338972568`

This keeps #199's semantic stop-delivery fix separate from the pre-existing running-operation data race it exposed.

xsachax:
Restacked onto the new #184 prerequisite after #232.

- base #184 head: `3e3988d21f6ac92eac95dbe0e1167b3ec53fa9ea`
- new #202 head: `7964fc623f0f16d13bb6461fe1d887846540bf04`
- duplicate atomic commit `04f7f9a` was explicitly dropped; the atomic publication now comes only from #232
- published ancestry: `f376590` → `3e3988d` → `7964fc6`
- old-to-new tree delta is limited to #232’s four files

Validated before the force-with-lease update with `go test -tags bingonative -race -count=1 ./...`, `go vet -tags bingonative ./...`, Linux/amd64 debugger-test cross-compilation, and `git diff --check`. #225 has been given this exact frozen head.

PR #203, opened 2026-08-09
fix(hub): own stable logical breakpoint ids

### LINKED ISSUE #200: [hub] Buffered pre-Restart clear can delete a different breakpoint
**Describe the bug**

A breakpoint clear generated before `Restart` can be injected after the replacement debugger is live. Because the hub exposes each engine's physical breakpoint IDs and a fresh engine restarts ID allocation from 1, that stale clear can alias a different breakpoint in the new process and physically remove the wrong trap.

The dangerous boundary is earlier than hub command injection:

1. DAP `setBreakpoints` generates `ClearBreakpoint{ID: oldPhysicalID}` and queues the marshalled command in `Handler.cmdOut`.
2. The hub client read-pump calls `Handler.ReadMessage`, but that command can remain buffered or blocked before `injectCommand`.
3. A second client injects and completes `Restart`. Breakpoints are compactly reinstalled in the new engine, reusing old numeric IDs for different locations.
4. The first read-pump finally injects the old clear. The hub passes the stale ID directly to the new debugger, which clears whichever new breakpoint now owns that number.

This is a P1 debugger-integrity defect: the target breakpoint remains armed, another breakpoint silently disappears, and the DAP adapter's transactional FIFO records the opposite physical result.

An epoch stamped in `Hub.injectCommand` is not a fix. By injection time, Restart is already complete, so the stale command is stamped with the new epoch even though its ID was bound to the old engine when the Handler generated it.

**Code locations**

- `internal/dap/breakpoints.go` — `advanceLineLocked` marshals a clear with the currently installed debugger ID and queues it in `cmdOut`.
- `internal/dap/handler.go:283-296` — `Handler.ReadMessage` drains `cmdOut`.
- `internal/hub/client.go:117-133` — the per-client read-pump injects only after `ReadMessage` returns; other clients' read-pumps can win the race.
- `internal/hub/hub.go:402-429,460-495` — set/clear bookkeeping exposes and stores physical debugger IDs.
- `internal/hub/hub.go:540-575` — Restart creates a fresh engine, reinstalls in deterministic order, and replaces old IDs with the new engine's compact IDs.
- `internal/debugger/breakpoint.go:33-89` — every fresh breakpoint table allocates IDs from 1 and clears by ID.

**To Reproduce**

1. Use a real `hub.Hub` and real DAP `Handler` with a faithful debugger fake whose breakpoint table allocates IDs from 1 on each new engine.
2. Install A and B with a deliberate ID gap (for example, A=id2 and B=id3).
3. Send a DAP replace-all request that removes A. Wrap the Handler in a gating `WSConn`: let `ReadMessage` obtain the `ClearBreakpoint{id2}` bytes, then block before returning them to the hub read-pump.
4. Through a second hub client, execute Restart to completion. The fresh debugger reinstalls A=id1 and B=id2.
5. Release the held clear.
6. Observe physical state: A remains armed, B is disarmed. The Handler consumes the clear confirmation positionally as A and believes A was removed.

This real Hub + real Handler boundary test reproduces deterministically under `-race`. A negative control that releases the clear before Restart is safe: A is removed first and only B is reinstalled.

**Expected behavior**

Breakpoint IDs visible on the session protocol must remain stable logical identities across debugger replacement. The hub should own logical IDs and translate them to the active engine's physical IDs for Set, Clear, BreakpointHit, and Restart bookkeeping. Restart must update only the logical-to-physical mapping. Then a delayed clear for logical A either clears A in the replacement engine or is rejected if A was discarded; it can never alias B.

Logical IDs should remain monotonic for the lifetime of the managed session, including a later fresh Launch/Attach, so a stale command from a prior target cannot alias a newly-created breakpoint. Unknown or failed clears must retain/report state consistently rather than deleting mapping optimistically.

**Screenshots**

N/A; the deterministic boundary test asserts the replacement debugger's physical table and the Handler's logical state.

 - OS: Darwin 24 (reproduction uses in-process hub/DAP fakes under `-race`)
 - Architecture: arm64; behavior is platform-independent
 - Version: `0509b53bfc44da889e152ee20f56ce3663882c65` plus the DAP restart/transaction correction stack

**Additional context**

Two adversarial reviews confirmed the race. The final review constructed the real Hub + real Handler + gated-`ReadMessage` reproduction and specifically disproved injection-time generation stamping.

#127 / draft PR #138 reconciles DAP state from `RestartedPayload`, and #178 / draft PR #182 makes `setBreakpoints` transactional. Both are prerequisites for truthful adapter state, but neither prevents the hub from sending a stale physical ID to the replacement engine. The fix should therefore be a separate layer above PR #182.

A weaker alternative is globally non-reused physical IDs across replacement engines. Hub-owned session-stable logical IDs are preferred because they preserve the identity clients already hold and protect every client type, not only DAP.

xsachax:
Fixes #200

## Stack

`#138` → `#182` → **this PR**

Based on `xsachax-fix-dap-breakpoint-transactions` (#182). Review that first — this PR builds on its breakpoint transaction rather than replacing it.

Rebased onto the frozen #182 head `7a30a7148d8b617810c6f73d77af659e9d11314a`. The restack was a pure replay — each of this PR's three commits kept an identical `git patch-id`, so the only tree change relative to the previous head is #182's own test commit.

## The defect

`breakpointTable.nextID` (`internal/debugger/breakpoint.go`) restarts at 1 in every engine, so a **physical** breakpoint id only means anything to the engine that issued it — and `handleRestart` replaces that engine.

A command can be *generated* against one engine and *injected* after it has been replaced. The DAP adapter marshals a `ClearBreakpoint{ID: oldPhysicalID}` into `Handler.cmdOut`; the hub read pump can be descheduled between `Handler.ReadMessage` returning those bytes and `Client.readPump` calling `injectCommand` (`internal/hub/client.go`). Another client completes a Restart in that window. The stale id then names a **different** breakpoint in the fresh process, and releasing it disarms the wrong trap.

Reproduced deterministically under `-race` with a real `Hub` + real DAP `Handler` and a gating `WSConn`: installed A=phys2 / B=phys3, held `clear(2)`, restarted to A=phys1 / B=phys2, released the clear, and observed **A armed / B deleted** while the Handler believed A was cleared.

## Why an epoch does not fix it

An epoch stamped at `Hub.injectCommand` is *provably insufficient*: provenance is bound when the Handler **generates** the bytes, but injection happens after the Restart and would read the **new** epoch. The command is indistinguishable from a fresh one at the point it is stamped.

## The fix

The hub owns breakpoint identity. Client-visible ids are session-stable **logical** ids, translated to per-engine physical ids at every boundary. New `internal/hub/breakpoints.go`:

- **Monotonic for the whole hub lifetime.** `reset()` (a fresh `Launch`/`Attach`) drops the mappings but deliberately does **not** rewind the high-water mark — re-minting is exactly what would let a delayed command from a previous target alias a breakpoint in the new one.
- **`Set`** installs, mints a logical id, and broadcasts `EventBreakpointSet` with it. The engine's id never leaves the hub.
- **`Clear`** rejects an unknown logical id **without touching the debugger** (the load-bearing half), translates a known one to the active physical id, and only `untrack`s after the debugger confirms. **No optimistic deletion** — a rejected clear leaves the trap armed (`breakpointTable.clear` keeps its entry when the memory write fails), so the client must keep being able to name it.
- **`Restart`** iterates logical ids deterministically, reinstalls each location, re-homes the mapping onto the new physical id, and reports `RestartedPayload.Breakpoints` under the **same** logical ids. Entries that fail to resolve are removed and reported as `Discarded`; a later clear naming one is rejected safely.
- **`BreakpointHit`** physical ids are rewritten to logical before broadcast, on both broadcast paths in `handleEvent`.
- **A hit can race its own clear.** The engine emits into a buffered channel and `engine.ClearBreakpoint` has no suspend guard, so `Run`'s `select` may execute a queued clear before draining a hit generated earlier. Removals are therefore *retired* per engine (bounded, dropped by `reset()`) so the late hit still reports the id the client held.
- **Adopted ≠ installed.** A physical id that is neither live nor retired is adopted for reporting rather than passed through (pass-through could collide with a logical id naming a different breakpoint — this keeps raw `hub.New` hubs coherent), but it is **not** a restart target: the hub never armed it and must not arm it on the replacement process.

All five protocol surfaces carrying a breakpoint id are covered: `BreakpointSetPayload.Breakpoint.ID`, `BreakpointClearedPayload.ID`, `BreakpointHitPayload.Breakpoint.ID`, `RestartedPayload.Breakpoints[].ID`, and the inbound `ClearBreakpointPayload.ID`.

`CmdSetBreakpoint`/`CmdClearBreakpoint` are removed from the generic `dispatch` switch and routed to hub-aware handlers, preserving centralized error wrapping and exactly one confirmation per command. No changes to `internal/debugger`. No correlation ids, no epochs, no DAP-only hacks.

### Follow-up: a cleared breakpoint the engine resurrects

Semantic review found a second, independent way a logical id could be stranded — not an aliasing bug, but its mirror image.

Clearing the breakpoint the tracee is **parked on** is undone by the engine. `resumeFromBreakpoint` replays the `e.lastBP` entry pointer it stashed at hit time rather than re-reading the table, and `breakpointTable.reinstall` re-adds that entry under its **original physical id**. The hub had already retired the mapping on the confirmed clear, so the resurrected trap was reported under a logical id that `CmdClearBreakpoint` then rejected — and the location could not be re-set either, because the engine still had the address armed (`errBreakpointExists`). The result was a phantom breakpoint that was permanently neither removable nor settable.

The retirement record is now **bidirectional**:

- `retiredLogical` (physical → logical) reports a late or resurrected hit under the id the client holds.
- `retiredPhysical` (logical → physical) keeps that reported id **clearable**.

A tombstone-clear is deliberately **not** consumed: the tracee may still be parked there and resurrect it again, so the id has to stay reportable and removable. This cannot weaken the #200 fix, because `breakpointTable.set` never reuses a physical id within one engine, and `reset()` purges the record on every `Launch`/`Attach`/**and Restart** — a tombstone can only ever name the breakpoint it was created for, never one in another generation. A dedicated spec pins that by restarting into an engine which reuses the retired physical id and asserting the replacement engine never sees a `ClearBreakpoint`.

### Out-of-band Launch (documented, not a regression)

A *second* client issuing a fresh `CmdLaunch`/`CmdAttach` on a shared session invalidates every logical id: mappings reset while the high-water mark advances. A client still holding old ids gets a clean `EventError` rather than silently disarming a stranger's breakpoint — which is exactly what the pre-fix raw physical ids did, so this is a strict improvement. It is **not** transparent, though: a DAP adapter whose `bpLine`s still record pre-Launch ids believes lines are armed that the new process lacks, and only finds out when a removal fails. `CmdRestart` — what the DAP `restart` request maps to — is the supported way to relaunch *while preserving* ids. An explicit invalidation event would be a new protocol surface and is deliberately out of scope for #200, which is about not corrupting the ids that do survive. Recorded in `AGENTS.md`.

## Tests

- **Permanent regression** (`internal/dap/hublogicalbp_test.go`): real `Hub` + real DAP `Handler` + a gating `WSConn` at the exact `Handler.ReadMessage` → `Client.readPump` boundary. Holds a generated clear, completes a Restart via a second client, releases the clear, and asserts the **physical** new-engine A is cleared, B remains armed, and the Handler's logical state agrees.
- **Negative control**: the same clear injected *before* the Restart behaves normally, so the fix does not change the in-order path. It stays green under the bug-restoring mutation, which is what a control should do — but that is a sanity check on the mutation's blast radius, not evidence about what the headline test isolates. (An earlier revision of this description claimed the latter; that inference was invalid.)
- **Multi-client**: WebSocket and DAP observers see the same stable ids. Restarts first so logical and physical genuinely diverge, and asserts `physicalA != logicalA` before checking translation, so the test cannot pass vacuously.
- **Hub unit specs** (12, `internal/hub/breakpoints_test.go`): logical id broadcast, clear translation, unknown-id rejection, failed-clear retention, hit translation, restart preservation while physical ids compact, discarded-mapping rejection, monotonicity across a fresh Launch, hit-racing-its-own-clear (extended: the reported id must still be clearable), never re-arming a breakpoint the hub did not install, keeping a step-off-resurrected breakpoint clearable, and refusing to let a retired id survive into a replacement engine. The fake engine allocates from 101 so any leaked physical id is immediately visible.

### Mutation checks

| Mutation | Caught by |
|---|---|
| Expose physical ids (old behavior) | `TestStaleClearAcrossRestartHitsTheRightBreakpoint` — `armed lines [10] (want [20])`, i.e. the literal #200 defect — plus 6 hub specs |
| `reset()` also rewinds the high-water mark | `does not re-mint logical ids after a fresh Launch` |
| Skip `BreakpointHit` translation | 3 hub specs + `TestWebSocketAndDAPSeeTheSameIDs` |
| Drop the retirement lookup | `reports a hit that raced its own clear under the id the client held` |
| Adopted mappings become restart targets | `never re-arms a breakpoint the hub did not install` |
| Drop the `retiredPhysical` clear fallback | `keeps a step-off-resurrected breakpoint clearable` |
| `reset()` forgets to purge the tombstone | `does not let a retired id survive into a replacement engine` |

## Validation

- `gofmt` / `goimports` clean; `go vet -tags bingonative ./...` clean; `golangci-lint run --build-tags bingonative ./...` — **0 issues**
- `go test -tags bingonative -race ./internal/hub ./internal/dap` — 5× repeated, all green
- `go test -tags bingonative -race ./...` — all packages green
- **Native darwin/arm64 E2E** (`restart || fullstack || dap || breakpoints`) — 8/8 through the whole stack
- Diff contains only this layer relative to #182 head

## Compatibility

- **Wire version unchanged (1.2).** Breakpoint ids were already opaque `int`s; only their ownership and lifetime changed. No field shape or payload change.
- **The one observable behavior change:** a `ClearBreakpoint` naming an id the hub does not know now returns `EventError` instead of forwarding to the debugger. That is the fix — previously such a command could silently disarm an unrelated breakpoint.
- Ids are now stable across Restart (they previously changed silently), which is strictly more useful to clients.
- The DAP transaction from #182 is unmodified. Its positional `setQ`/`clearQ` FIFOs now receive logical confirmations that are consistent with the actual physical effect, and `reconcileRestartBreakpointsLocked`'s id adoption becomes a no-op since logical ids survive a restart.
- Raw `hub.New` behavior is preserved via adoption; no e2e or hub test depends on raw id pass-through.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1368,9 +1368,24 @@ no-breakpoint target, not a clear-then-continue.
 id (`installedID`). After `EventRestarted`, reconcile its exact source-path/line
 keys from `RestartedPayload` before replying: retained entries adopt the fresh
 debugger id, while discarded entries are removed and emit a `breakpoint` changed
-event with their prior DAP id and `verified:false`. A line with an op in flight
-is skipped — that op still owns the line and its confirmation is still queued.
-Do not reset `setQ`/`clearQ`; those FIFOs still own any in-flight confirmations.
+event with their prior DAP id and `verified:false`. A line with an operation in
+flight is re-identified **too** — correlation rides the queued `*bpOp`, not
+`installedID`, so adopting the fresh id cannot desynchronise the FIFOs, whereas
+keeping the pre-restart id would strand the line on an identifier the new engine
+never issued (the already-marshalled clear fails, retain-on-failure reissues that
+dead id forever, and the real new breakpoint becomes unremovable). Only a line
+with `installedID == 0` is skipped: it owns no identity yet, so a payload entry
+for it describes another driver's breakpoint. Do not reset `setQ`/`clearQ`; those
+FIFOs still own any in-flight confirmations.
+
+**This transaction depends on reliable ordinary-command delivery** (#160 / #162).
+There are deliberately no adapter-side timeouts — inventing one would mask
+command loss and could fire against a merely slow debugger. A silently dropped
+`SetBreakpoint`/`ClearBreakpoint` therefore produces no confirmation, which now
+latches that line's `pending` forever: no further operation can be issued for it
+and the owning request is never answered. The adapter side of the contract is
+covered here (it blocks rather than drops, and wire order matches slot-reservation
+order); the hub side is `injectCommand`'s admission.
 
 ### Server wiring + multi-client discovery
 
@@ -1513,8 +1528,9 @@ translator keeps DAP entirely outside the hub — a strictly additive package.
   breakpoint transaction: a rejected clear retains the id and fails its request,
   partial clear failure, a superseded pending set clears exactly once, an
   overlapping pending set issues one command, set→remove→re-add leaves one live
-  breakpoint, cross-source independence, restart around an in-flight op,
-  exactly-once responses). Run with the normal
+  breakpoint, cross-source independence, restart re-identification with an
+  in-flight operation, no-drop/in-order command delivery, exactly-once
+  responses). Run with the normal
   `go test -tags bingonative ./internal/dap/...`.
 - E2E: label `dap` in [test/integration](test/integration/) — a real go-dap
   client over TCP through the WHOLE stack (client → TCP →
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1328,8 +1328,18 @@ legitimately disagree:
   `EventBreakpointSet` and dropped **only** by a confirmed
   `EventBreakpointCleared`. Never on intent.
 - `desired` — the newest request's intent for that line (latest wins).
-- `pending` — the single in-flight `bpOp` for that line. One op per line at a
+- `pending` — the single LIVE `bpOp` for that line. One live op per line at a
   time is what keeps the id-less `setQ`/`clearQ` FIFOs unambiguous.
+
+An operation is **abandoned** when a restart discards its line: it stops being
+that line's `pending` op (so the line can converge again immediately instead of
+latching behind a removal that already happened), but it keeps its place in the
+FIFO, because the debugger will still answer it and that answer must pop the head
+it reserved or every later confirmation correlates to the wrong request.
+`liveLineLocked` is the gate — it matches the popped op against the line's
+current `pending`, so an abandoned (or superseded, or forgotten) op's answer does
+exactly one thing: pop. It never settles waiters, fails owners, or writes state
+that now belongs to a re-identified line.
 
 `advanceLineLocked` is the convergence loop: it issues the one command the
 installed/desired gap calls for, or settles everyone parked on the line when the
@@ -1343,6 +1353,9 @@ two already agree. Every confirmation re-enters it, which yields the invariants:
   leave a breakpoint the client can never remove — `breakpointTable.clear` keeps
   the entry when the memory write fails, so the trap really is still armed. The
   failure path deliberately does not re-enter the loop; a new request retries.
+  That no-retry rule is justified **only** while the line is still armed: with
+  nothing armed there is nothing to retry and nothing to report, so the removal
+  completes successfully and the line resumes converging rather than stalling.
 - A **pending Set superseded before it confirms** is cancelled on confirm: the id
   is recorded as a fact (the trap IS armed) but `desired` stays false, so the
   next advance clears it immediately. It is never committed as desired state.
@@ -1375,8 +1388,12 @@ keeping the pre-restart id would strand the line on an identifier the new engine
 never issued (the already-marshalled clear fails, retain-on-failure reissues that
 dead id forever, and the real new breakpoint becomes unremovable). Only a line
 with `installedID == 0` is skipped: it owns no identity yet, so a payload entry
-for it describes another driver's breakpoint. Do not reset `setQ`/`clearQ`; those
-FIFOs still own any in-flight confirmations.
+for it describes another driver's breakpoint. A **discarded** line goes further:
+it abandons its in-flight operation (see above) — the command was addressed to a
+breakpoint the relaunched process no longer has, so its answer can only be stale,
+and leaving it live would latch the line behind a removal that already happened.
+Do not reset `setQ`/`clearQ`; those FIFOs still own any in-flight confirmations,
+abandoned ones included.
 
 **This transaction depends on reliable ordinary-command delivery** (#160 / #162).
 There are deliberately no adapter-side timeouts — inventing one would mask
@@ -1529,8 +1546,9 @@ translator keeps DAP entirely outside the hub — a strictly additive package.
   partial clear failure, a superseded pending set clears exactly once, an
   overlapping pending set issues one command, set→remove→re-add leaves one live
   breakpoint, cross-source independence, restart re-identification with an
-  in-flight operation, no-drop/in-order command delivery, exactly-once
-  responses). Run with the normal
+  in-flight operation, a discarded line abandoning its operation (later set,
+  already-satisfied removal, late stale success), no-drop/in-order command
+  delivery, exactly-once responses). Run with the normal
   `go test -tags bingonative ./internal/dap/...`.
 - E2E: label `dap` in [test/integration](test/integration/) — a real go-dap
   client over TCP through the WHOLE stack (client → TCP →
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1318,18 +1318,59 @@ connection on sustained overload. Never replace overload eviction with an
 uncorrelated bingo error event — the adapter cannot identify a dropped tail
 slot, so such an event could consume the wrong FIFO head.
 
-**setBreakpoints is replace-all** (`breakpoints.go`): diff the requested lines
-for a source against `bpByFile` — clear removed, set new, keep unchanged — and
-respond once every slot in the request resolves, in request order. Clearing the
-breakpoint the process is currently parked on is supported by the engine and
-pinned by the native `breakpoints` acceptance test on both platforms.
+**setBreakpoints is a replace-all TRANSACTION** (`breakpoints.go`): a request
+records the source's latest-wins intent and is answered only once every operation
+it owns has been confirmed. `bpByFile` is `file → line → *bpLine`, and `bpLine`
+deliberately keeps three things apart, because pipelined requests make them
+legitimately disagree:
 
-`bpByFile` keeps the stable DAP id separate from the debugger's internal id.
-After `EventRestarted`, reconcile its exact source-path/line keys from
-`RestartedPayload` before replying: retained entries adopt the fresh debugger
-id, while discarded entries are removed and emit a `breakpoint` changed event
-with their prior DAP id and `verified:false`. Do not reset `setQ`/`clearQ`;
-those FIFOs still own any in-flight confirmations.
+- `installedID` — an installed FACT. Written **only** by a confirmed
+  `EventBreakpointSet` and dropped **only** by a confirmed
+  `EventBreakpointCleared`. Never on intent.
+- `desired` — the newest request's intent for that line (latest wins).
+- `pending` — the single in-flight `bpOp` for that line. One op per line at a
+  time is what keeps the id-less `setQ`/`clearQ` FIFOs unambiguous.
+
+`advanceLineLocked` is the convergence loop: it issues the one command the
+installed/desired gap calls for, or settles everyone parked on the line when the
+two already agree. Every confirmation re-enters it, which yields the invariants:
+
+- A **removal is owned by the request that caused it** (`openClears` +
+  `clearOwners`), so a response never precedes its own clears. A remove-all no
+  longer reports success while the `ClearBreakpoint` is still in flight.
+- A **rejected clear retains the mapping** and fails the originating request
+  (`clearFailure` → an error `setBreakpoints` response). Forgetting the id would
+  leave a breakpoint the client can never remove — `breakpointTable.clear` keeps
+  the entry when the memory write fails, so the trap really is still armed. The
+  failure path deliberately does not re-enter the loop; a new request retries.
+- A **pending Set superseded before it confirms** is cancelled on confirm: the id
+  is recorded as a fact (the trap IS armed) but `desired` stays false, so the
+  next advance clears it immediately. It is never committed as desired state.
+- An **overlapping request wanting an already-pending line attaches** to that op
+  (another `setWaiter`) instead of issuing a duplicate `CmdSetBreakpoint`, which
+  the engine would reject with `errBreakpointExists` — and whose failure handling
+  used to delete the entry holding the live id.
+- **Exactly one response per request** (`bpRequest.ready()` claims the right to
+  answer; slots and obligations settle on different goroutines), and requests are
+  answered in `reqSeq` order so a later request's view lands last.
+- Sources are independent; the id-less FIFO limitation is unchanged.
+
+Breakpoint commands are staged in `outbox` and drained by `flushCommands`
+(serialised by `flushMu`) rather than enqueued directly: both the DAP read loop
+and the hub write pump produce them, and the FIFOs only correlate if the wire
+order matches the order slots were reserved under `mu`.
+
+Clearing the breakpoint the process is currently parked on re-arms it through the
+engine's step-off path (see the clearbp spec), so the e2e continue-to-exit uses a
+no-breakpoint target, not a clear-then-continue.
+
+`bpLine` keeps the stable DAP id (`dapID`) separate from the debugger's internal
+id (`installedID`). After `EventRestarted`, reconcile its exact source-path/line
+keys from `RestartedPayload` before replying: retained entries adopt the fresh
+debugger id, while discarded entries are removed and emit a `breakpoint` changed
+event with their prior DAP id and `verified:false`. A line with an op in flight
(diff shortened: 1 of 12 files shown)


xsachax:
## BLOCKING, in-PR: M9 — the "no optimistic deletion" invariant is not test-enforced

**This is not a post-merge follow-up.** It is a **blocking, in-PR requirement** for #203: the test-only strengthening below **must** land in this PR, as a **fourth, test-only commit**, during the authorized final restack (onto the **exact then-current `main` at #203's queue turn — after #182, #162, and #176 have merged**; see criterion 1), immediately after the three existing commits are replayed patch-identically — and **before exact-head acceptance**.

It is deliberately **not** applied to the currently frozen head `d06d0d100496d4fcdf0bb56d9272bfc8cf20d33f`, which remains the accepted frozen source until that restack is authorized. The patch artifact is preserved byte-exact in the meantime (`sha256:b8d54662e5ec0f023725587c46b433f35f693cc6fdbf91646a897574169aad58`, verified to `git apply --check` cleanly against `d06d0d1`).

### The gap

`AGENTS.md` and `internal/hub/breakpoints.go` both document a load-bearing invariant:

> **No optimistic deletion:** a rejected clear leaves the trap armed (`breakpointTable.clear` keeps its entry when the memory write fails), so the client must keep being able to name it.

That invariant is currently **not enforced by any test**. Moving `h.bps.untrack(logicalID)` from *after* the debugger confirms the removal to *before* it — i.e. deleting the mapping optimistically — leaves the entire suite green.

The reason is the bidirectional tombstone added in commit `d06d0d1`. `untrack` retires the pair into `retiredPhysical`, so `physicalForClear` still resolves a dropped id and a retry clear still succeeds. The existing spec, *"retains the mapping when the debugger rejects a clear"* (`internal/hub/breakpoints_test.go` L282–300), asserts only that the id remains **nameable** — which the tombstone guarantees either way. It is therefore **non-discriminating**.

### Why it matters (not behaviour-neutral)

Retention has to mean the breakpoint is still **live**, not merely still nameable. With optimistic untrack the mapping leaves `byLogical`, so `installedLogical()` excludes it, and `sortedRestartTargets()` therefore omits it. Consequence: a breakpoint the debugger **refused to remove** is still armed on the old process, but is silently **absent from the relaunched one** after `CmdRestart`.

Only a relaunch distinguishes the two behaviours. That is precisely what the proposed spec asserts.

### Mutation evidence (re-verified on the frozen head, `d06d0d1`)

Mutant **M9** — optimistic untrack in `Hub.clearBreakpoint`:

```go
// mutant: untrack BEFORE the debugger confirms
h.bps.untrack(logicalID)
if err := h.dbg.ClearBreakpoint(physicalID); err != nil {
    return dispatchResult{}, err
}
```

| run | tree | result |
|---|---|---|
| 1 | frozen head, **as-is** + M9 mutant | `./internal/hub/...` and `./internal/dap/...` **both green** — mutant SURVIVES |
| 2 | frozen head + patch below, no mutant | `./internal/hub/...` **green** (53/53) — new spec passes clean |
| 3 | frozen head + patch below + M9 mutant | `./internal/hub/...` **FAIL — 52 Passed, 1 Failed** |
| 4 | frozen head + patch below + M9 mutant | `./internal/dap/...` green (mutant is invisible to DAP) |

Run 3 fails **specifically at the Restart-reinstallation assertion**, as required:

```
Summarizing 1 Failure:
  [FAIL] logical breakpoint ids [It] keeps a rejected clear's breakpoint a restart target
  internal/hub/breakpoints_test.go:321

  [FAILED] a retained breakpoint must still be a restart target
  Expected
      <[]protocol.Breakpoint | len:0, cap:0>: nil
  to have length 1

Ran 53 of 53 Specs
FAIL! -- 52 Passed | 1 Failed | 0 Pending | 0 Skipped
```

Note that in run 3 the **pre-existing** retention spec still passes — the new spec is the sole discriminator, which is the direct proof that the old one was non-discriminating.

All commands were `go test -tags bingonative -race -count=1`. All mutations were reverted; the worktree is clean and the head remains exactly `d06d0d1`.

### The patch (apply verbatim as the 4th commit)

Verified to apply cleanly to `d06d0d1` (`git apply --check` → OK). Test-only; touches no production code.

```diff
diff --git a/internal/hub/breakpoints_test.go b/internal/hub/breakpoints_test.go
index 90ae469..37eaf85 100644
--- a/internal/hub/breakpoints_test.go
+++ b/internal/hub/breakpoints_test.go
@@ -299,6 +299,32 @@ var _ = Describe("logical breakpoint ids", func() {
 		Expect(fleet.at(0).armedLines()).To(BeEmpty())
 	})
 
+	// Retention has to mean the breakpoint is still LIVE, not merely still
+	// nameable. The tombstone keeps even a dropped id clearable, so a retry
+	// succeeding cannot distinguish the two — only a relaunch can: a breakpoint
+	// the debugger refused to remove is still armed, so the hub must reinstall
+	// it. Dropping the mapping before the debugger confirms would leave the trap
+	// armed on the old process and silently absent from the new one.
+	It("keeps a rejected clear's breakpoint a restart target", func() {
+		startManaged(nil)
+
+		logical := setBP(conn, "main.go", 10)
+		fleet.at(0).setClearErr(fmt.Errorf("restore bytes failed"))
+		conn.inject(mustCommand(protocol.CmdClearBreakpoint,
+			protocol.ClearBreakpointPayload{ID: logical}))
+		waitForEventKind(conn, protocol.EventError, nil)
+
+		conn.inject(mustCommand(protocol.CmdRestart, protocol.RestartPayload{}))
+		var restarted protocol.RestartedPayload
+		waitForEventKind(conn, protocol.EventRestarted, &restarted)
+
+		Expect(restarted.Breakpoints).To(HaveLen(1),
+			"a retained breakpoint must still be a restart target")
+		Expect(restarted.Breakpoints[0].ID).To(Equal(logical))
+		Expect(fleet.at(1).armedLines()).To(Equal([]int{10}),
+			"the relaunched process must carry the breakpoint that was never removed")
+	})
+
 	It("translates a BreakpointHit's physical id to the logical id", func() {
 		startManaged(nil)
 
```

### Acceptance criteria for the restack (all blocking, all in this PR)

1. **Replay parent is the exact then-current `main` at #203's queue turn (after #182, #162, and #176), not a branch head.** The stack queue order is **#181 → #182 → #162 → #176 → #203**. #203 must **not** be rebased directly onto a `#182` head, and this must **not** be read as "`main` immediately after #182" — the later preceding layers **#162** and **#176** merge first. Replay only once all three (#182, #162, #176) have merged, onto whatever `main` is at that exact moment. Replay the three reviewed commits patch-identically (`git patch-id` equality per commit).
2. Add the patch above as a **fourth, test-only commit in this PR**, before exact-head acceptance.
3. The new spec must prove **both** halves of retention:
   - the rejected clear's breakpoint is still **logically installed** — it appears in `RestartedPayload.Breakpoints` under the *same* logical id, which is exactly the `installedLogical()` → `sortedRestartTargets()` path; and
   - it is **actually reinstalled on Restart** — the relaunched engine carries the armed line (`fleet.at(1).armedLines()`).
4. Re-run the no-optimistic-deletion (M9) mutant and require it to fail **specifically** at the Restart-reinstallation assertion (`breakpoints_test.go` — "a retained breakpoint must still be a restart target"), not merely somewhere in the suite.
5. `go vet -tags bingonative ./...`, full `go test -tags bingonative -race -count=1 ./...`, and a **genuine full `just e2e-darwin` pass in which every spec in the then-current suite passes** — zero failures, zero unexplained skips — before labeling/merge. The suite count is deliberately **not** hard-coded: it must be whatever the suite contains at restack time, not a historical number. A partial pass is **not** acceptable as final; the base reproduction of the pre-existing darwin `churn` failure establishes non-causality for triage only and does **not** waive the repository-wide gate.
6. CAS push uses `--force-with-lease=refs/heads/xsachax-stabilize-breakpoint-ids:d06d0d100496d4fcdf0bb56d9272bfc8cf20d33f`; the lease stays pinned to `d06d0d1` unless the direct refs drift, in which case stop and re-snapshot rather than overwriting.
7. **Explicitly** retarget this PR's base ref to **`main`** — owned by this PR, not done centrally — and keep it a **draft**. Ordering is fixed: only **after #176 has landed** and **after the CAS-published head has been verified** (live ref == PR head == CI `head_sha`). Do **not** rely on GitHub auto-retargeting when #182's branch disappears: a deleted upstream branch can silently retarget this PR to that branch's own base rather than `main`. Closure requires a live read of `baseRefName == main` — verify it, do not assume it. Not to be done before the restack is authorized.

### Other mutation survivors (non-blocking, recorded for completeness)

- **M1** — `retire()`'s duplicate-physical guard. Unreachable today: physical ids are never reused within one engine. Defensive.
- **M2** — `dropPhysical`'s `cur == logical` guard. Unreachable today: `Restart` calls `reset()` before re-binding, so two logical ids never point at one physical id. Defensive.
- **M3** — `retiredCap` (1024) eviction. A pure memory bound; observing it needs >1024 retirements within a single engine. Low severity.

Killed (genuinely covered): raw-physical-id exposure, clear-translation bypass, both `localizeBreakpointIDs` broadcast paths, adopt-not-installed, pass-through-instead-of-adopt, tombstone-survives-reset, the `retiredPhysical` fallback, high-water-mark rewind, restart re-homing `bind`, and restart ordering.

> **Caveat for anyone re-running these:** naive mutant deletions can produce **compile** failures that masquerade as kills (e.g. removing `sort.Ints` orphans the `sort` import). Always distinguish a compile error from a real test failure.

PR #210, opened 2026-08-09
fix(wsmon): truncate one-line output on code-point boundaries

### LINKED ISSUE #209: [wsmon] Output truncation can split UTF-8 code points
**Describe the bug**
`cmd/wsmon`'s `oneLine` helper measures a Go string in bytes and truncates it with `s[:117]`. If byte 117 falls inside a multibyte UTF-8 code point, wsmon emits invalid UTF-8 followed by `...`. Terminals render a replacement glyph, and CJK/emoji text is truncated much earlier than the intended display width.

The helper renders arbitrary tracee stdout/stderr, panic messages, and snapshot rejection messages, so non-ASCII text is valid end-user data.

Priority: **P3 / low**.

**To Reproduce**
1. Construct output with 116 ASCII bytes followed by Japanese text and enough trailing bytes to exceed the truncation threshold.
2. Pass it through the same operation as `oneLine`: replace newlines, then take `s[:117] + "..."`.
3. Observe the result ends with the leading byte of the Japanese rune and is invalid UTF-8.
4. Run a target that emits this text and view it in wsmon; observe a replacement glyph.

**Expected behavior**
Truncated output must remain valid UTF-8 and must never split a code point. The display budget should count complete runes (or explicitly back off to a valid UTF-8 boundary) before appending `...`.

**Screenshots**
N/A.

 - OS: all supported platforms
 - Architecture: linux/amd64 and darwin/arm64
 - Version: `main` at `0509b53bfc44da889e152ee20f56ce3663882c65`

**Additional context**
Code locations:

- `cmd/wsmon/main.go:176-188`: arbitrary event text reaches renderer state.
- `cmd/wsmon/main.go:403-411`: `oneLine` uses byte length and byte slicing.

Two independent reviews reproduced the exact byte boundary and confirmed the path is reachable with valid user output. The synthetic parent-cycle renderer candidate from the same audit was rejected because valid runtime parent-goid data cannot create that cycle.

A focused fix should truncate by complete runes or use `utf8.RuneStart`/`utf8.ValidString` to back off safely, with table tests for ASCII boundaries, Japanese text, emoji, combining sequences, embedded newlines, and already-short input. This touches `cmd/wsmon` code already changed by the snapshot semantics PR, so implement it as a dependent layer rather than against stale `main`.

xsachax:
## Summary

`cmd/wsmon`'s `oneLine` measured its budget in bytes and cut with `s[:117]`. The helper renders arbitrary tracee stdout/stderr, panic text, and snapshot rejection messages, so when byte 117 landed inside a multi-byte rune wsmon emitted **invalid UTF-8** before the ellipsis and terminals drew a replacement glyph. The same byte budget also truncated CJK and emoji at a third to a quarter of the intended length (40 emoji = 160 bytes was cut mid-rune despite being far inside a sane one-line budget).

The budget now counts **complete code points**.

Fixes #209

## Stack

`#148 → #192 → this PR`

This PR targets `xsachax-fix-snapshot-request-semantics` (#192), not `main`. #192 rewrote the surrounding `cmd/wsmon` code (fire-and-forget snapshot requests, `-once` bounding, `-timeout` validation, custom usage), so #209 has to land as a dependent layer rather than against stale `main` — the issue says as much. Review only the top commit; merge after #192.

## Design decisions

Two were deliberate and are worth confirming:

**1. Rune budget, not a byte budget with boundary back-off.** The issue sanctions either. Runes were chosen because they give a uniform character budget across scripts, are the closest proxy to on-screen length that needs *no* display-width dependency, and are byte-for-byte identical to the old behaviour for pure ASCII (the common case: `RuneCountInString(ascii) == len(ascii)`, and the rune cut walks one byte per iteration, so it returns exactly `s[:117]`).

The accepted trade-off, documented in the code: 120 wide runes can exceed 120 terminal *columns*. Measuring columns needs an East Asian width table, which is explicitly out of scope. Output stays bounded (≤ 120 runes, ≤ ~471 bytes).

**2. Input that is already invalid UTF-8 passes through untouched.** Each invalid byte counts as exactly one rune, matching `utf8.RuneCountInString` and Go's `range` semantics — the stdlib's `RuneCountInString` *is* `for range s { n++ }`, so the count and the cut use the identical iterator and cannot disagree, including for truncated sequences, overlongs, and surrogates. `oneLine` truncates; it does not sanitize. Rewriting bytes to U+FFFD would change output on the *non*-truncating path too, which is unrelated to #209. The precise invariant held is therefore: **truncation never introduces invalidity the input did not already contain**, and it is asserted directly in the tests.

Consequence of not adding segmentation: a grapheme cluster (base + combining mark, ZWJ emoji sequence) can still be split at the cut. Valid UTF-8 either way, documented, and pinned by tests.

## Scope

Surgical and deliberately narrow:

- Newline normalization is **byte-for-byte unchanged** (`\n` → the two-character escape; a bare `\r` still survives, which is pre-existing and out of scope — the `embedded crlf` case characterizes it so a future CR fix has to update it consciously).
- No display-width, grapheme-segmentation, or any other new dependency.
- No other rendering changes; no other call site depends on `oneLine`'s byte length.
- `AGENTS.md` needs no update — no documented invariant changes.

## Validation

- `go test -tags bingonative -race -count=1 ./cmd/wsmon/...` — pass
- `go test -tags bingonative -race -count=1 ./pkg/client/... ./pkg/protocol/... ./cmd/...` — pass (parent #192's wsmon and client suites included, no regressions)
- `go vet -tags bingonative ./...` — clean
- `go build -tags bingonative ./...` — clean
- `gofmt -l .` / `goimports` — clean (lefthook pre-commit ran)

**Test coverage.** A table over short / exact / over-budget boundaries: empty and whitespace-only input, short ASCII, surrounding-space trim, embedded LF, embedded CRLF, trailing LF, ASCII at 120 and at 121, the issue's exact repro (116 ASCII + Japanese, where byte 117 lands on a leading byte), Japanese at and over budget, astral emoji under and over budget, a ZWJ sequence at the cut, combining marks separated from their base, a cut landing inside the newline escape, short invalid UTF-8, and an invalid byte before the cut. Every case additionally asserts the output is within budget and that valid input never produces invalid output. A sweep test runs every byte prefix of a mixed 1/2/3/4-byte string through `oneLine` and asserts validity throughout, and `truncateRunes` is pinned directly (`n == 0`, shorter than `n`, exactly `n`, boundary cut, invalid byte).

**Mutation-checked.** Reverting the cut to `s[:oneLineKeepRunes]` fails 6 table cases plus the sweep. Independently, reverting only the *budget* to `len(s)` while keeping the rune-safe cut fails `japanese at budget` and `emoji under budget` — so the tests pin the design decision, not merely UTF-8 validity, and would also catch the issue's alternative "byte budget with back-off" implementation.

Independently reviewed; no issues found. Reviewer separately re-derived the four tricky test expectations (newline escape, combining marks, invalid byte, ZWJ) and confirmed the off-by-one behaviour of `truncateRunes`.

Draft until #192 merges.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -245,6 +245,23 @@ automatic — is delivered exactly once on `Events()`, where a client treats the
 uniformly as telemetry. It must stay out of the correlated set: never route it
 through `sendAndWait`. See issue #187.
 
+Two consequences a caller must handle, both inherent to an id-less broadcast
+protocol rather than to this change:
+
+- **Delivery is best-effort.** The answer rides the shared `Events()` buffer,
+  which `readPump` drops from when the consumer stops draining, and a rejected
+  request answers with `EventError`(`CmdGoroutineSnapshot`) instead of a
+  snapshot. Anything that blocks waiting for a snapshot must handle both — see
+  `cmd/wsmon`'s `-once`, which fails fast on a rejection or a closed stream
+  rather than waiting forever.
+- **The answer is broadcast.** A query is fanned out to every client on the
+  session like any other event, and it carries empty `Created`/`Exited`. An
+  observer that renders "the latest snapshot's deltas" (as `cmd/wsmon`'s
+  lifecycle panel does) therefore blanks that panel until the next automatic
+  stop; an append-only timeline (the VS Code view) is unaffected. Making a query
+  addressable to its requester needs wire-level correlation and is deliberately
+  out of scope here — a candidate for a telemetry 1.3.
+
 `syncMu` serializes synchronous commands, but a timeout does **not** cancel the
 command already sent to the server. The pending queue therefore retains a
 timed-out request as retired reply debt until its matching confirmation or
@@ -706,6 +723,16 @@ consumed the pending deltas — the following automatic snapshot diffed against
 the query, so goroutines created and exited in between appeared in neither
 (issue #187). Do not add a delta-bearing query path back: the wire payload is
 unchanged, and a client that wants deltas must read the automatic stream.
+
+The regression gate is the `concurrency`-labelled `declareBaselineOwnershipSpec`
+E2E, and it is deliberately shaped: a query at an ordinary breakpoint stop sees
+exactly the live set that stop's automatic snapshot already adopted, so adopting
+again would be a no-op and prove nothing. The spec instead **steps over a `go`
+statement** — steps are the one suspend that carries no automatic snapshot — so
+the query observes a goroutine the baseline has never seen, then asserts the
+NEXT automatic snapshot still reports it as created. It fails if the query is
+made lifecycle-tracking or the automatic path non-tracking; the unit tests
+around `snapshotFrom` pin the seam's semantics but cannot catch a rewiring.
 
 **Graceful fallback.** Every read is best-effort. `resolveGoLayout` marks the
 layout invalid if any required `g`/`gobuf`/`stack`/`m` offset is missing, and any
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -219,12 +219,31 @@ In [pkg/client](pkg/client/), the `Client` interface splits methods by what
 they wait for:
 
 - **Synchronous** (`Restart`, `SetBreakpoint`, `ClearBreakpoint`, `Locals`,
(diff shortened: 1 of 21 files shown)


xsachax:

Issue #194, opened 2026-08-09
[telemetry] Unbounded goroutine events kill the VS Code concurrency observer


xsachax:
**Describe the bug**

The two goroutine-carrying events — `EventGoroutineSnapshot` and `EventGoroutines` — are produced with no size bound, while the VS Code concurrency observer enforces hard bounds on both the WebSocket transport and its own decoder. On any realistically-concurrent target the producer outruns both limits, and the observer does not merely drop a frame: it treats the failure as a transport problem and burns its entire reconnect ladder, then terminates the Bingo Concurrency view for the rest of the debug session.

There are two independent, deterministic failure paths. Both are producer-side contract gaps, not consumer bugs, and both are permanent for the session once triggered.

---

**To Reproduce**

*Reproduction 1 — `EventGoroutineSnapshot` exceeds the WebSocket transport limit and is never delivered*

1. Start `bingo` with `-dap-addr`, launch any target that reaches a few thousand live goroutines, and attach the VS Code companion so the Bingo Concurrency view joins the session over `/ws?session=<id>`.
2. Hit a breakpoint. `internal/debugger/goroutines.go` `goroutineSnapshot()` walks `runtime.allgs` up to `maxGoroutineScan = 8192` and returns every goroutine unchanged; `internal/hub/dispatcher.go` marshals it into one `EventGoroutineSnapshot` frame with no size check.
3. `editors/vscode/src/observer.ts` constructs its `ws` socket with `maxPayload: maximumEnvelopeBytes` (2 MiB). `ws` rejects the oversized frame in the transport layer with `WS_ERR_UNSUPPORTED_MESSAGE_LENGTH` and closes with 1006.

Observed: the message is **never delivered to `decodeEvent`** — the failure happens below the decoder, so no amount of consumer-side validation can see it. `onError` fires, `onClose` fires, and `#reconnect` walks the full `reconnectDelays` ladder. The observer connects 1 + 6 times, each new connection immediately requesting a fresh snapshot that is oversized for exactly the same reason, and then latches `connection: "error"` with `telemetry reconnect limit reached`. The view is dead for the rest of the session, and every reconnect costs a full snapshot round trip on the server.

[...]

PR #220, opened 2026-08-09
fix(telemetry): close the gaps that still kill the observer

xsachax:
**Part of #194.** Stack: **#149 → #196 → this PR.** Targets `xsachax-telemetry-bounded-goroutine-events` (PR #196).

> **Rebased.** This branch was originally written as Layer B against `3cd7d43`, when #196 was still the dormant-contract-only layer. #196 has since grown its own complete producer wiring, version bump, extension 0.4.0 and consumer activation. Rather than compete with it, this branch was rebased onto the current parent head (`775edb9`) and **reduced to only what the parent does not have**: the review-driven correctness fixes and the test assets that were missing. It no longer redesigns the producer.

## Why these are not polish

The parent puts the contract in force. These are the paths that still kill the Concurrency view or tell the reader something untrue.

### 1. A legal server broadcast could permanently kill the observer

`validatePayload` applies the bounded family's caps (string length, array length, node budget, depth) to **every consumed kind** — but most of those kinds are ones the contract explicitly leaves unbounded. Every rejection was a `TelemetryProtocolError`, which the observer treats as *proven and deterministic*: it latches, closes the socket and suppresses the reconnect ladder for the session.

Reachable with no malicious actor:

1. Paste a >4080-character expression into a Watch or the Debug Console.
2. `internal/dap/requests.go` forwards it verbatim; `dwarf.go` returns `no variable named %q in scope`.
3. The hub broadcasts `EventError` with that message to **every** client, including the telemetry observer.
4. `boundedString` rejects it → fatal → the Bingo Concurrency view is dead for the rest of the session.

`walkLimits`/`tooLarge` now classify by whether the kind is bounded. Structural failures (wrong type, unknown key, bad enum) and the bounded kinds stay terminal; size overruns elsewhere are ordinary errors, so the ladder still applies and the view recovers on the next snapshot.

```
Error msg 5000 chars              transient   (ladder intact, view recovers)
BreakpointHit fn name 5000        transient
Stepped 6000 frames               transient
Paused 21000 nodes                transient
GoroutineSnapshot status 5000     FATAL       (bounded — a retry replays it)
SessionState bad enum             FATAL       (structural)
```

`bounded` is **latent** today and deliberately so: neither bounded kind reaches that walk (`GoroutineSnapshot` goes to the typed decoders, `Goroutines` is shallow-ignored), so the bounded family's fatality comes from those decoders. It exists so adding either kind to `consumedKinds` cannot silently downgrade a real violation to a retry. The comment says exactly that, so nobody mistakes it for the live protection.

### 2. Refresh — the documented recovery — did not work from every terminal state

It only cleared the fatal latch. Exhausting the reconnect ladder is equally terminal and never latches: it leaves `#socket === undefined`, so `refresh()` fell through to `#sendSnapshot()`, which returns immediately with no socket. Silent no-op, no recovery, dead panel. It now redials whenever nothing is left to ask.

The panel also rendered **"Connecting to telemetry"** while permanently stopped — it looks busy, so nobody presses the one control that would recover it. It now names the state it is actually in (`Telemetry disconnected` / `Reconnecting` / `Telemetry connection closed`) and, on error, points at the error rather than promising a retry will succeed.

### 3. `readThreads` called a truncated count exact

Clipping was reported only for the `maxThreadScan` ceiling. A walk that ended on an unreadable `alllink` also left threads unvisited — which is precisely what makes the count a lower bound. Reporting otherwise is the dishonesty the totals exist to prevent.

### 4. The rendering said things it could not know

- An empty tree blamed *"The target may be exiting or runtime inspection degraded"* even when `Totals` proved the debugger had 41,203 goroutines — and a **filter that matched nothing** got the same message, sending someone to hunt a debugger problem that does not exist. Now attributed to the cause the evidence supports: filter / omission / unreadable runtime.
- Threads reported one shortfall **twice** (empty-state *and* omission note). Each collection is now explained once, beside its own data.
- `wsmon` captioned a degraded read `(complete)` — a degraded snapshot omits nothing, so it wore the caption — and printed `goroutines=0 threads=0 (complete)` before any snapshot arrived. It now certifies nothing; the `included/total` form is the signal, and its absence needs no caption.
- Clipped wording contradicted itself (`(8192 live)` beside "that total is a lower bound"). Now `of at least 8192 found` / `the debugger stopped after finding 8192 goroutines, so more may exist`, consistently across the webview, `wsmon` and `cli`.
- Summary cards showed `0 Goroutines` before the first snapshot — a measurement nobody has taken. Now `—`.

### 5. The producer/consumer string limit was asserted twice, never proven once

The producer counts UTF-16 code units in Go; the consumer counts them in JavaScript. Both suites tested that rule *in their own terms*, which is exactly how they drift — and a disagreement at the boundary is a deterministic kill. The 13 boundary cases now live in **one generated fixture both read** (`pkg/protocol/testdata/goroutine_string_boundary.json`). A Go spec generates and diff-gates it (`BINGO_UPDATE_FIXTURES=1`); the VS Code suite asserts the identical verdict on the identical wire strings, **and** that Node's replacement-mode UTF-8 decode of the raw bytes matches what `encoding/json` produced.

Covers ASCII / astral (2048 code points = 4096 units) / 4096 astral code points / 3-byte BMP / raw `0xFF` / lone-surrogate bytes / U+2028–9, each at the limit and one unit over.

### 6. The producer wiring had no direct coverage

`internal/debugger/goroutines_producer_test.go` drives the **real** runtime reader — DWARF-resolved offsets against a synthetic `runtime.allgs`/`allm` planted in the fake backend — and pins what the parent's wiring promises: independent scan-clipped flags (ceiling *and* unreadable-link), events bounded on both ceilings, `Current` and its ancestor chain surviving as anchors, a flat list that truncates rather than emptying (an empty threads response makes the DAP translator invent a synthetic `main`), lifecycle deltas diffed against the runtime rather than the trimmed frame, and the degraded fallback not touching `prevGoids`.

## Verification

```
gofmt, go vet -tags bingonative ./...            clean
go test -tags bingonative -race -count=1 ./...   all packages ok
just vscode-check                                244 tests, 0 fail
                                                 (lint, typecheck, bundle, list)
npm run test:integration (pinned Electron)       exit 0
just vscode-package                              reproducible two-build gate:
                                                 identical binary + VSIX hashes,
                                                 archive + 0.4.0 verified
darwin/arm64 native E2E, all labels except churn 21/21 pass
```

### `churn` — pre-existing on the parent, not from this branch

The darwin `churn` spec fails on this branch **and on the parent**, at the target's ~180s watchdog:

| binary | runs | result |
| --- | --- | --- |
| parent `775edb9` | 3 | 1 pass (171.7s), 2 fail (181.3s, 181.2s) |
| this branch | 4 | 4 fail (180.8s) |

Both sit on the same boundary. This branch adds **no code to the e2e path** — its only non-test Go change is `readThreads`' return expression, and the new `_test.go` files are in `internal/debugger`/`pkg/protocol`, which are not linked into `test/integration`'s binary. AGENTS.md documents this exact hazard: extra per-stop work in the snapshot pushes `churn` past its target's watchdog. Worth a look on the parent — the packer now marshals every element on every breakpoint snapshot.

## Reviews

Two independent adversarial reviews were run against the pre-rebase branch and every finding fixed; the fixes are what this branch now carries.

- **Code review** found §1 (High) and verified the fix: no remaining path from a legal broadcast to a `TelemetryProtocolError`, the bounded family not weakened, and `refresh()` race-free with `#connectionEpoch`/`dispose()` (no double-connect while a reconnect delay is pending).
- **UX review** found §2, §4 and the singular-grammar issue, then re-reviewed and found the filter mis-attribution and the duplicate thread note — both fixed here.

## Known limitation

`wsmon`/`cli` still cannot *label* the degraded synthetic-goroutine snapshot as such — the wire carries no flag for it (the webview uses a heuristic). They no longer overclaim, but they cannot affirmatively say "this is the fallback". Adding a flag is a protocol change outside this contract's scope.

## Scope

No generic hub or `Location` cap, no client read cap, no chunking, no compression, no producer redesign, no unrelated features.


xsachax:
Superseded by the consolidated telemetry implementation in #196; closing this divergent activation fork to keep one authoritative stack.

xsachax:
Superseded — closing unmerged.

This branch was written as the activation layer against the old parent head `3cd7d43`, when #196 was still the dormant-contract-only layer. #196 has since advanced through activation and its own review fixes (`b7f136b` → `865d07d`): both producers wired, wire 1.3, extension 0.4.0, decoder/observer activation, per-collection clipped flags, the clipped fast path, degraded packing, CLI/SDK totals, and docs.

Ancestry confirms the divergence rather than a lag:

- fork point `3cd7d43`; head `41ff20b` (tree-identical to `92d197b`, which was amended only to trigger CI)
- `git merge-base --is-ancestor 41ff20b <#196 head>` → **false**

So this PR now duplicates #194 from a materially different, older tree, and its producer design competes with the one #196 landed rather than building on it. #196's approach is the one to keep — notably `Debugger.Goroutines()` returning the payload while `client.Client` keeps its slice API plus a separate totals accessor, which has a smaller blast radius than the interface change here.

Nothing is lost by closing. Every part of this branch that #196 does not contain was rebased onto #196's current head and now lives in #223:

| Not in #196 | Rehomed in #223 |
| --- | --- |
| Decoder fatality scoped for the payload walk, not just the byte check | ✅ |
| `refresh()` redialling after an exhausted reconnect ladder | ✅ |
| Terminal-state headline instead of a permanent "Connecting" | ✅ |
| Empty tree attributed to a filter vs an omission vs an unreadable runtime | ✅ |
| Per-collection omission reported once, beside its own data | ✅ |
| Cross-language UTF-16/invalid-UTF-8 boundary fixture | ✅ |
| Producer specs over a synthetic `runtime.allgs`/`allm` | ✅ |
| `readThreads` reporting a walk that broke on an unreadable `alllink` | ✅ |

Review #196 and #223; this branch needs no further attention.

PR #223, opened 2026-08-09
fix(telemetry): close the observer, reconnect and honesty gaps above the contract

xsachax:
**Part of #194. Draft.**

Restacked onto exact `origin/main` `be2996d2f2270f824f6fb5d44144601c338ac03b`.
The former prerequisites (#219, #238, #224, and #196) are now on `main`; none
of their commits are replayed here.

## History

The accepted six-commit layer is preserved in order:

| Previous | Restacked |
| --- | --- |
| `9bf67bd` | `fadf887` |
| `7b1379e` | `9d9ebd0` |
| `225a066` | `c9cac00` |
| `3493e83` | `ed5d37a` |
| `e2da671` | `b9b0919` |
| `3536b76` | `6c53f04` |

Two explicit restack integration commits follow:

- `8f09efc` bumps the shipped VS Code extension from 0.4.0 to 0.4.1, as required
  by the installed-runtime upgrade boundary.
- `2395b4f` binds the raw numeric-overflow test frame to
  `wireProtocolVersion`; the accepted test still used a historical literal
  `"1.3"` and otherwise failed before reaching the classification under the
  landed wire 1.4 contract.

The old body referred to squashing historical commits `083c6ef` and `0d242b9`.
That remove/restore pair is already absent from the accepted six-commit layer,
and its final partial-thread-walk behavior is present on `main` through #196.
No absent commits were invented or replayed.

Preservation proof:

- all six old/new stable patch IDs match;
- normalized change-only cumulative diffs are byte-identical
  (`sha256:3229824edc1fec0734e692b2cf4b3019b16e79811b095de1b4f1a6c67c3ea33d`);
- range-diff is five `=` entries; the first differs only in base context
  (`protocol 1.3` wording on the historical base versus `1.4` on current main)
  while retaining the same stable patch ID;
- `main..HEAD` is exactly the six accepted commits plus the two integration
  commits, with merge-base exactly `be2996d2`.

### Bounded-family error classification

Size, depth, node, object-width, field-name, and string overruns are classified
by the family whose wire contract owns the limit:

- bounded telemetry kinds remain fatal, because retrying replays a proven
  contract violation;
- contract-legal unbounded consumed kinds close and enter the reconnect ladder
  instead of permanently latching the view;
- structural schema, type, and enum violations remain fatal where their owning
  validator proves them.

The generic walk's `bounded` flag remains deliberately latent: typed
`GoroutineSnapshot` decoding and shallow `Goroutines` handling keep both bounded
kinds out of that walk, and a dedicated test guards that precondition.

### Recovery

Refresh now recovers both terminal observer shapes: a fatal protocol latch and
an exhausted session-cumulative reconnect ladder. It redials whenever no socket
remains, restores the full ladder, and preserves the connection-epoch guard that
prevents a stale pending backoff from opening a second socket.

### Truthful rendering

Terminal states no longer claim to be connecting. Empty/filter/omission copy
distinguishes what is known, reports goroutine and thread shortfalls beside the
correct collection exactly once, and leaves pre-snapshot counts unknown instead
of fabricating zeroes.

### Shared producer/consumer boundary

One generated UTF-16 boundary fixture is consumed independently by Go and
TypeScript, covering ASCII, astral characters, three-byte BMP characters,
invalid UTF-8, lone-surrogate bytes, and line separators at and beyond the
limit. Current main retains #196's producer packing/totals/current-anchor specs,
including independent lower-bound reporting for a partial unreadable `allm`
walk; this follow-up does not alter those producer semantics.

## Version and scope

- VS Code extension: **0.4.1**
- wire protocol: **1.4 unchanged**
- no packer contract, totals/lifecycle/current-anchor behavior, degraded flags,
  DAP ID-0 semantics, snapshot-query behavior, generic hub cap, client read cap,
  chunking, or compression changes

Known limitations deliberately remain:

- the automatic reconnect ladder is session-cumulative and is not reset by a
  successful handshake; manual Refresh restores it after exhaustion;
- Refresh does not cancel a socket stuck indefinitely in `CONNECTING`.

## Exact-head verification

Validated at `2395b4fe75f6900b231a44616a5a94cb110498df`:

- gofmt + `go tool goimports`
- targeted tagged race suites:
  `./internal/debugger ./pkg/protocol ./cmd/wsmon`
- full `go test -tags bingonative -race -count=1 ./...`
- `go vet -tags 'bingonative e2e' ./test/integration`
- Linux-context `golangci-lint` (`0 issues`)
- Linux/amd64 server and E2E cross-builds (both verified ELF x86-64)
- shared fixture regeneration and Go/TypeScript drift gates
- `just vscode-check`: **258/258 tests**, including the real-WebSocket transport,
  exhausted-ladder Refresh, stale-backoff epoch, bounded-family mutation gates,
  UTF-16 fixture, and webview DOM/security gates
- pinned VS Code Electron integration 1.107.1: exit 0
- reproducible signed Darwin package:
  - binary SHA-256 `39cecb5957ba92bad5fed8367acb4c392d96d3f354fbc98c6e4b587cb023b89d`
  - VSIX SHA-256 `09cc26b898b111e9c572b9f4a0dfd052447cbd908159639b8be5f7b2054cb838`
  - package contents/version/signature verification passed
- packaged graphical E2E: all five progressive examples passed and the managed
  server exited on idle
- signed Darwin telemetry labels: **3 passed**, one intentional heavy
  current-goroutine proof skipped
- full exact-head `just e2e-darwin`: **24 passed, 1 intentional heavy skip**
- independent code-review and rubber-duck passes: no substantive findings

Exact-head GitHub status after publication:

- trusted Darwin verification gate: pass
- SonarCloud: pass
- normal Go, Debugger E2E, and VS Code workflow matrix: absent (GitHub created
  no runs for this SHA after the PR was retargeted from its obsolete stacked
  base), so there was no remote matrix to wait on


xsachax:
## Handoff note for the final restack owner — read before restacking

This PR is **frozen** pending the coordinated #196/#219/#224 composition. Two things you need that are not obvious from the diff.

### 1. The tip is `0d242b9`, not `607eb25`

Coordination notes have repeatedly referred to `607eb25`. That commit is real, but two later commits sit on top of it:

- `083c6ef` — removed the `readThreads` fix (instructed: don't duplicate #196's lower-bound semantics)
- `0d242b9` — restored it (instructed: `readThreads` is this PR's to keep; only the *degraded* flags are #196's)

**They cancel.** `git diff 607eb25 0d242b9` is **+2 lines** — one added assertion. Squashing that pair out during the restack is safe and returns this to a clean four-commit follow-up. The semantic weakening that must be corrected still lives at `0d242b9`, in `internal/debugger/goroutines_producer_test.go`.

### 2. The mandated correction, pre-verified and mutation-gated

Apply this **after** merging the accepted #196 base — it asserts #196's degraded semantics (`9839178`), so it fails on any base that predates them. It also closes a mutant that survived the first campaign.

<details>
<summary><code>composed-contract-correction.patch</code> — one file, 55 lines</summary>

```diff
diff --git a/internal/debugger/goroutines_producer_test.go b/internal/debugger/goroutines_producer_test.go
index 4ce2120..90bbe75 100644
--- a/internal/debugger/goroutines_producer_test.go
+++ b/internal/debugger/goroutines_producer_test.go
@@ -351,18 +351,45 @@ var _ = Describe("goroutine event production", func() {
 
 			snap, err := d.GoroutineSnapshot()
 			Expect(err).NotTo(HaveOccurred())
-			// The documented contract: the stop is still reported, as the legacy
-			// single synthetic goroutine, rather than erroring. What Totals says
-			// about a snapshot that carries no real runtime data is a property of
-			// the packer, covered by its own suite, and deliberately not pinned
-			// from here.
+			// The composed contract. The stop is still reported as the legacy
+			// single synthetic goroutine rather than erroring, AND the counts
+			// that stand for a runtime nobody could read are marked as the floor
+			// they are: absent totals mean "complete and exact" on the wire, so
+			// emitting none here would claim the runtime holds exactly one
+			// goroutine and no threads. "At least one" is the honest reading
+			// whether the runtime is merely not up yet or a scan failed partway.
 			Expect(snap.Goroutines).To(HaveLen(1))
 			Expect(snap.Goroutines[0].ID).To(Equal(1))
 			Expect(snap.Goroutines[0].Status).To(Equal("waiting"))
+			Expect(snap.Totals).NotTo(BeNil(),
+				"a degraded read must not present itself as a complete, exact count")
+			Expect(snap.Totals.Goroutines).To(Equal(1))
+			Expect(snap.Totals.Threads).To(Equal(0))
+			Expect(snap.Totals.GoroutinesClipped).To(BeTrue())
+			Expect(snap.Totals.ThreadsClipped).To(BeTrue())
 			Expect(snap.Created).To(BeEmpty(), "a degraded read invents no lifecycle events")
 			Expect(snap.Exited).To(BeEmpty())
 		})
 
+		It("marks the degraded goroutine LIST as a lower bound too", func() {
+			// The snapshot is not the only degraded shape on the wire: the
+			// Goroutines query answers a DAP threads request from the same
+			// synthetic stand-in, and absent totals there make the same claim of
+			// an exact count. Both degraded paths must read as a floor, or the
+			// honesty depends on which request the client happened to send.
+			img.seedThreads(4)
+			img.unreadableRuntime(true)
+
+			payload, err := d.Goroutines()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(payload.Goroutines).To(HaveLen(1))
+			Expect(payload.Goroutines[0].ID).To(Equal(1))
+			Expect(payload.Totals).NotTo(BeNil(),
+				"a degraded list must not present itself as a complete, exact count")
+			Expect(payload.Totals.Goroutines).To(Equal(1))
+			Expect(payload.Totals.GoroutinesClipped).To(BeTrue())
+		})
+
 		It("does not let an unreadable runtime look like every goroutine exited", func() {
 			img.seedGoroutines(8, 8, 3)
 			img.seedThreads(4)
```

</details>

**Re-gate on the accepted base — all three must fail:**

| Mutant | Edit | Expected |
| --- | --- | --- |
| A | degraded snapshot → `PackSnapshot(..., false, false)` | FAIL |
| B | degraded **list** → `gs, clipped = …, false` | FAIL |
| C | `readThreads` → `return out, i == maxThreadScan && mptr != 0` | FAIL |

**B is the one worth keeping.** It survived the original campaign because it mutates the degraded path of `readGoroutines`, not `goroutineSnapshot` — a path the snapshot spec cannot see. It answers the DAP `threads` request from the same synthetic stand-in, so without the second spec a snapshot client is told "at least one" while a DAP client is told "exactly one", and the honesty depends on which request the client happened to send. The patch adds that spec; B then fails.

Verified locally: merging #196's then-current tip into `0d242b9` **auto-resolves** — #196's degraded `true, true` and this PR's `readThreads` `mptr != 0` coexist without conflict.

xsachax:
## Second prepared correction — reconnect-budget hardening (not applied; PR is frozen)

An independent verification session drove this PR's `refresh()` fix against a discriminator probe and confirmed it correct on every path, including two properties stronger than the original report assumed: refresh restores a **full fresh 6-slot ladder**, and refresh during a pending backoff does not double-dial (the `#connectionEpoch` bump makes the sleeping `#reconnect` abandon its dial).

It also found a real weakness **in this PR's test**, which I reproduced.

### The gap

`observer.test.ts`'s refresh spec asserts `connection === "connecting"` after recovery. That pins `attempt === 0` only *incidentally* — the label is computed from the parameter (`attempt === 0 ? "connecting" : "reconnecting"`). A plausible future refactor, moving the ladder index into instance state and forgetting to zero it on refresh, keeps the parameter at `0` and **survives**:

```ts
#ladder = 0;                                    // new instance field
this.#ladder = Math.max(this.#ladder, attempt); // in #connect
void this.#reconnect(this.#ladder + 1, epoch);  // in onClose
```

Reproduced locally: with that mutation applied, `observer.test.ts` passes **7/7**. Refresh still redials, still reports `connecting`, still connects, still sends a snapshot — but the recovered connection has **zero** budget and goes terminal on the next blip. Refresh becomes single-use, which is exactly the failure the spec exists to prevent.

### The fix, verified

Assert the **budget**, not the label: after the recovered socket opens, drive one more disconnect and require it to re-enter the ladder at its floor.

| State | Result |
| --- | --- |
| clean code + hardening | **7/7 pass** |
| ladder-in-instance-state mutation | **fails** (was 7/7 before) |
| original M1 (`refresh()` gated on `#fatal` only) | **still fails** |

The harness change is one line — record the requested `milliseconds` instead of discarding it — so the assertion can pin the 100 ms floor directly rather than inferring a fresh ladder from a label.

<details>
<summary><code>observer-budget-hardening.patch</code> — one file, 54 lines</summary>

```diff
diff --git a/editors/vscode/test/observer.test.ts b/editors/vscode/test/observer.test.ts
index d5ad214..21f902f 100644
--- a/editors/vscode/test/observer.test.ts
+++ b/editors/vscode/test/observer.test.ts
@@ -53,19 +53,19 @@ class FakeSocket extends EventEmitter implements Socket {
 function setup(): {
   readonly observer: TelemetryObserver;
   readonly sockets: FakeSocket[];
-  readonly delays: { resolve: () => void }[];
+  readonly delays: { resolve: () => void; milliseconds: number }[];
 } {
   const sockets: FakeSocket[] = [];
-  const delays: { resolve: () => void }[] = [];
+  const delays: { resolve: () => void; milliseconds: number }[] = [];
   const dependencies: ObserverDependencies = {
     createSocket() {
       const socket = new FakeSocket();
       sockets.push(socket);
       return socket;
     },
-    delay(_milliseconds, signal) {
+    delay(milliseconds, signal) {
       return new Promise((resolve, reject) => {
-        const delay = { resolve };
+        const delay = { resolve, milliseconds };
         delays.push(delay);
         signal.addEventListener("abort", () => {
           reject(new Error("cancelled"));
@@ -206,6 +206,25 @@ describe("telemetry observer", () => {
       1,
       "a recovered connection asks for a snapshot so the view repopulates",
     );
+
+    // Recovery must restore the BUDGET, not merely dial once. Asserting the
+    // label above would pass a regression that redials with the ladder still
+    // spent, making Refresh single-use: the panel comes back and dies on the
+    // next blip, which is the failure this spec exists to prevent.
+    const spent = delays.length;
+    sockets[exhausted]!.close();
+    assert.notEqual(
+      observer.model.connection,
+      "error",
+      "a recovered connection must not be one blip from terminal",
+    );
+    assert.equal(observer.model.connection, "reconnecting");
+    assert.equal(delays.length, spent + 1, "the disconnect must enter the ladder");
+    assert.equal(
+      delays[spent]!.milliseconds,
+      100,
+      "the recovered ladder restarts at its floor rather than resuming where it stopped",
+    );
     observer.dispose();
   });
 
```

</details>

**Apply at restack alongside `composed-contract-correction.patch`** (posted above), then re-run both gates on the final base.

### Two residuals, explicitly out of scope for this PR

- The **automatic** ladder still accumulates across independent healthy incidents: six provably healthy connect/observe cycles are survived, the seventh is terminal. This PR does not change that — it converts the outcome from *permanently stranded* to *recoverable by one Refresh with a full reset*, which lowers the severity of that separate defect but does not fix it.
- `refresh()` is gated on `#socket === undefined`, so it cannot break out of a socket hung in `CONNECTING` (peer accepts TCP, never completes the handshake). There is no connect timeout. Narrow, self-heals if the peer closes.

xsachax:
## Supersedes the previous hardening comment — now covers a second surviving mutant

The verification session ran one further mutation, **M4**: drop the `#connectionEpoch` staleness check in `#reconnect` (`if (!this.#disposed && this.#connectionEpoch === epoch)` → `if (!this.#disposed)`), so a sleeping backoff no longer abandons its dial.

**Reproduced: my suite passed 7/7 under M4.** Nothing in this PR — or anywhere in the repo — pinned that guard, yet the regression it models is real: a Refresh issued during a pending backoff races a second live socket into `#socket` and leaks the loser.

The budget spec cannot catch it, and for a structural reason: it drives a ladder that is *already exhausted*, so no backoff is ever in flight when Refresh runs.

So the prepared patch now carries **two** specs, replacing `observer-budget-hardening.patch`:

1. **Recovered-budget assertion** — after Refresh recovers, one more disconnect must re-enter the ladder at its 100 ms floor. Kills the ladder-in-instance-state mutant that made Refresh single-use while still reporting `connecting`.
2. **No double-dial on interrupted backoff** — adapted from the verifier's D2 into this repo's existing `FakeSocket`/`setup()` harness, so it carries no external dependency.

### Mutation matrix — combined suite, 8 specs

| Mutation | Result |
| --- | --- |
| pristine | **8/8 pass** |
| M1 — `refresh()` gated on `#fatal` only | **killed** (2 fail) |
| M2 — `refresh()` → `#connect(1)` | **killed** |
| M3 — ladder index in instance state, not zeroed on refresh | **killed** |
| M4 — `#connectionEpoch` staleness guard dropped | **killed** |

M1 fails two specs because losing the redial breaks both recovery paths; the rest fail exactly one each, so no spec is doing double duty by accident.

<details>
<summary><code>observer-reconnect-hardening.patch</code> — one file, 79 lines</summary>

```diff
diff --git a/editors/vscode/test/observer.test.ts b/editors/vscode/test/observer.test.ts
index d5ad214..54813b7 100644
--- a/editors/vscode/test/observer.test.ts
+++ b/editors/vscode/test/observer.test.ts
@@ -53,19 +53,19 @@ class FakeSocket extends EventEmitter implements Socket {
 function setup(): {
   readonly observer: TelemetryObserver;
   readonly sockets: FakeSocket[];
-  readonly delays: { resolve: () => void }[];
+  readonly delays: { resolve: () => void; milliseconds: number }[];
 } {
   const sockets: FakeSocket[] = [];
-  const delays: { resolve: () => void }[] = [];
+  const delays: { resolve: () => void; milliseconds: number }[] = [];
   const dependencies: ObserverDependencies = {
     createSocket() {
       const socket = new FakeSocket();
       sockets.push(socket);
       return socket;
     },
-    delay(_milliseconds, signal) {
+    delay(milliseconds, signal) {
       return new Promise((resolve, reject) => {
-        const delay = { resolve };
+        const delay = { resolve, milliseconds };
         delays.push(delay);
         signal.addEventListener("abort", () => {
           reject(new Error("cancelled"));
@@ -206,6 +206,50 @@ describe("telemetry observer", () => {
       1,
       "a recovered connection asks for a snapshot so the view repopulates",
     );
+
+    // Recovery must restore the BUDGET, not merely dial once. Asserting the
+    // label above would pass a regression that redials with the ladder still
+    // spent, making Refresh single-use: the panel comes back and dies on the
+    // next blip, which is the failure this spec exists to prevent.
+    const spent = delays.length;
+    sockets[exhausted]!.close();
+    assert.notEqual(
+      observer.model.connection,
+      "error",
+      "a recovered connection must not be one blip from terminal",
+    );
+    assert.equal(observer.model.connection, "reconnecting");
+    assert.equal(delays.length, spent + 1, "the disconnect must enter the ladder");
+    assert.equal(
+      delays[spent]!.milliseconds,
+      100,
+      "the recovered ladder restarts at its floor rather than resuming where it stopped",
+    );
+    observer.dispose();
+  });
+
+  // Refresh bumps #connectionEpoch, and the sleeping #reconnect re-checks it
+  // after its delay. Without that check a refresh issued during a pending
+  // backoff races a second live socket into #socket and leaks the loser. The
+  // budget spec above cannot see this: it drives a ladder that has already been
+  // exhausted, so no backoff is ever in flight while refresh runs.
+  it("does not double-dial when refresh interrupts a pending backoff", async () => {
+    const { observer, sockets, delays } = setup();
+    observer.start();
+    sockets[0]!.open();
+    sockets[0]!.close();
+    assert.equal(delays.length, 1, "a backoff is pending");
+
+    observer.refresh();
+    assert.equal(sockets.length, 2, "refresh dials once");
+
+    delays[0]!.resolve();
+    await new Promise((resolve) => setImmediate(resolve));
+    assert.equal(
+      sockets.length,
+      2,
+      "the superseded backoff must abandon its dial rather than open a second socket",
+    );
     observer.dispose();
   });
 
```

</details>

### Independently confirmed against the advanced base

The verifier composed the restack that this PR cannot yet perform — #196 `7c0f5fb` plus this PR's consumer-only observer delta plus the hardening — and reports the observer delta and the patch both **apply cleanly**, with the hardened suite green on the composed tree. **The `DIRTY` status is not in the observer files**; it is in the producer material being dropped to #219.

Deliberately **not** landed, on the verifier's own advice: their D0 (ladder accumulation across healthy incidents) belongs with whoever fixes accumulation, and their D3 pins the hung-`CONNECTING` behaviour this PR has chosen not to fix — landing it would freeze a wart as a contract.

xsachax:
## Transfer bundle for #196's owner — reader-level specs + composed-contract correction

Per the final ownership ruling, **#223 becomes consumer-only**: all `internal/debugger` production *and* test hunks leave this PR. The reader-level specs pin **#196's** production contract, so they belong with it. Posting the bundle here so it is durable and can be taken up without depending on my session.

**Awaiting receipt confirmation from #196's owner.** I am holding these artifacts until then.

### What transfers

**1. `internal/debugger/goroutines_producer_test.go`** — new file, drops in whole. Drives the *real* DWARF-resolved reader against a synthetic `runtime.allgs`/`allm` planted in `fakeBackend` memory, so it exercises the packed producer end to end rather than mocking it. Ten specs:

| Area | Spec |
| --- | --- |
| scan clipping | neither collection clipped on a walk that completes |
| | flags only the goroutine scan when `runtime.allgs` outruns its ceiling |
| | flags the thread scan when the M list outruns its ceiling |
| bounding | saturating snapshot stays inside the wire contract (byte + both element caps; `Current` and its whole ancestor chain survive as anchors) |
| | saturating goroutine list stays inside the contract, and **truncates rather than emptying** — an empty threads response makes the DAP translator invent a synthetic `main` |
| lifecycle | deltas diff the runtime, not the trimmed frame |
| | deltas cannot overflow the contract (`maxGoroutineScan <= MaxLifecycleDeltaIDs`) |
| degraded | falls back to the synthetic goroutine, `Totals` present with `Goroutines=1`, `Threads=0`, **both** clipped flags true |
| | the degraded goroutine **list** is marked a lower bound too |
| | an unreadable runtime never looks like every goroutine exited (`prevGoids` untouched) |

**`composed-contract-correction.patch` is already applied inside that file** — the two degraded specs above are its content. The standalone patch is included only for provenance.

**Excluded, as ruled:** the partial-walk spec (`"flags only the thread scan when the M list becomes unreadable"`) and its `truncateThreadWalkAfter` helper — those follow the `readThreads` semantics to #219/#224. `unreadableRuntime()` is retained; the degraded specs need it.

**2. `export_test_seams.patch`** — adds `ExportedRuntimeVarAddr`, `ExportedGoOffsets`, and the `ExportedMaxGoroutineScan`/`ExportedMaxThreadScan` constants. Required: the harness resolves runtime struct offsets from DWARF by name rather than hardcoding them, exactly as the reader does, so it cannot drift across Go versions.

**3. `fakebackend_hook.patch`** — adds `failReadFrom`/`failReadTo` to `fakeBackend.ReadMemory`. Required for the degraded specs: an unreadable runtime is expressed as a **read failure**, not as zeroed memory. That distinction is load-bearing — "this address reads as empty" depends on where the image puts its globals, which differs between ELF and Mach-O, and an earlier version of these specs passed on darwin while failing on Linux CI for exactly that reason.

### Why the degraded assertion matters

Absent totals mean *complete and exact* on the wire. A degraded read stands in for a runtime nobody could scan, so emitting no totals claims the runtime holds exactly one goroutine and no threads — the overclaim the totals exist to prevent, applied to the case where least is known. These specs pin `9839178`'s behaviour so it cannot silently regress.

### Verification status

`gofmt` clean. The specs were green at `0d242b9` and the composed-contract correction was verified against a local composition of #196's then-tip, where three mutants were driven (degraded snapshot `false,false`; degraded list `false`; `readThreads` ceiling-only) and all failed as intended. **Re-run on the final base is #196's owner's call** — two of those mutants target #196's own code, so only its final tip proves them.

<details>
<summary>Bundle location</summary>

Also held at `transfer-to-196/` in this session's artifacts: `goroutines_producer_test.go` (correction pre-applied), `export_test_seams.patch`, `fakebackend_hook.patch`, `composed-contract-correction.patch`.

</details>

xsachax:
## Transfer manifest + full payload (durable) — for #196 owner receipt

Correcting my earlier bundle comment: it **described** the transfer but did not contain the payload, so it was not actually durable. The `.go` file and the seam patches existed only in my session directory. Inlining everything here so receipt can be acknowledged against fixed hashes and the material survives independently of any session.

### Manifest

| File | Bytes | SHA-256 |
| --- | --- | --- |
| `goroutines_producer_test.go` | 15585 | `c541909ff385eb9b92a89b12cf910fc150a2930f1221b47ffcee4ae9425e4a67` |
| `export_test_seams.patch` | 2229 | `f14965a3322fdb73be9883ade62b7957da2bc7230f0fa9d3411f04f24fcd6aad` |
| `fakebackend_hook.patch` | 1229 | `f7520369f3f84ec2fdf66e052c09c716495386b7f9bcbe428f74587b0a621c01` |
| `composed-contract-correction.patch` | 2833 | `96b71c5d85d8a7141028cbc02511018a867fb977c658aa909e1f57cf70a653a0` |

**Spec count: 10** `It(...)` blocks across 4 `Describe(...)` groups — *not* the "seven" some coordination notes cite. Seven were the original reader specs; the composed-contract correction added the degraded-list spec and expanded degraded coverage. All reader-level specs transfer **except** the partial-walk one (`"flags only the thread scan when the M list becomes unreadable"`) and its `truncateThreadWalkAfter` helper, which follow `readThreads` to #219/#224 and are already removed from the file below.

**Seams added by `export_test_seams.patch`** — exactly three:
- `func ExportedRuntimeVarAddr(d Debugger, name string) (uint64, bool)`
- `func ExportedGoOffsets(d Debugger) (map[string]int64, bool)`
- `const ExportedMaxGoroutineScan = maxGoroutineScan`, `ExportedMaxThreadScan = maxThreadScan`

**Hook added by `fakebackend_hook.patch`** — `failReadFrom` / `failReadTo` on `fakeBackend`, plus the range check in `ReadMemory` and the `fmt` import.

`composed-contract-correction.patch` is **already applied** inside the `.go` file; it ships separately for provenance only. Applying it again will fail.

### Specs carried

| Group | Spec |
| --- | --- |
| scan clipping | neither collection clipped on a walk that completes |
| | flags only the goroutine scan when `runtime.allgs` outruns its ceiling |
| | flags the thread scan when the M list outruns its ceiling |
| bounding | saturating snapshot stays inside the wire contract; `Current` + whole ancestor chain survive as anchors |
| | saturating goroutine list truncates rather than emptying — an empty threads response makes DAP invent a synthetic `main` |
| lifecycle accounting | deltas diff the runtime, not the trimmed frame |
| | deltas cannot overflow the contract (`maxGoroutineScan <= MaxLifecycleDeltaIDs`) |
| degraded snapshots | synthetic fallback carries `Totals` with `Goroutines=1`, `Threads=0`, **both** clipped flags true |
| | the degraded goroutine **list** is marked a lower bound too |
| | an unreadable runtime never looks like every goroutine exited (`prevGoids` untouched) |

<details>
<summary><code>internal/debugger/goroutines_producer_test.go</code> — new file, drops in whole</summary>

```go
package debugger_test

import (
	"encoding/binary"

	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"

	"github.com/bingosuite/bingo/internal/debugger"
	"github.com/bingosuite/bingo/pkg/protocol"
)

// These specs drive the REAL runtime reader — DWARF-resolved offsets against a
// synthetic runtime.allgs / runtime.allm planted in the fake backend's memory —
// so the producer wiring is exercised end to end rather than mocked. They are
// what proves the goroutine event contract is actually in force at the source:
// that every produced event is bounded, that each collection's scan-clipped
// flag is independent and truthful, and that lifecycle accounting still runs
// against the full runtime rather than the trimmed frame. See issue #194.

// runtimeImage lays out a synthetic Go runtime in the fake backend's address
// space. Addresses are arbitrary but disjoint; sizes are generous so adjacent
// structs never overlap regardless of the host Go version's real struct sizes.
const (
	gArrayBase    = uint64(0x1000_0000) // the []*g backing array
	gStructBase   = uint64(0x2000_0000) // fabricated g structs
	gStructSpan   = uint64(0x400)
	mStructBase   = uint64(0x4000_0000) // fabricated m structs
	mStructSpan   = uint64(0x400)
	mStructRegion = uint64(0x1000_0000) // the whole span reserved for m structs

	// Stack window of the goroutine the tracee is stopped in. The reader
	// identifies the current goroutine by SP containment, so only this one
	// brackets the seeded SP.
	currentStackLo = uint64(0x7000_0000)
	currentStackHi = uint64(0x7000_1000)
	currentSP      = uint64(0x7000_0800)

	gStatusRunning = 2 // _Grunning
)

type runtimeImage struct {
	fb      *fakeBackend
	off     map[string]int64
	allgs   uint64
	allm    uint64
	current int
}

func newRuntimeImage(fb *fakeBackend, d debugger.Debugger) *runtimeImage {
	off, ok := debugger.ExportedGoOffsets(d)
	Expect(ok).To(BeTrue(), "the fixture binary must carry runtime DWARF")
	allgs, ok := debugger.ExportedRuntimeVarAddr(d, "runtime.allgs")
	Expect(ok).To(BeTrue())
	allm, ok := debugger.ExportedRuntimeVarAddr(d, "runtime.allm")
	Expect(ok).To(BeTrue())
	img := &runtimeImage{fb: fb, off: off, allgs: allgs, allm: allm}
	// Seeding one header must never disturb the other. Their real addresses come
	// from the image, whose global layout differs between ELF and Mach-O, so this
	// is asserted rather than assumed — an overlap would corrupt a spec in a way
	// that looks like a debugger bug.
	Expect(allgs+24 <= allm || allm+8 <= allgs).To(BeTrue(),
		"runtime.allgs (0x%x) and runtime.allm (0x%x) overlap", allgs, allm)
	// Start from an explicitly empty runtime rather than from whatever unseeded
	// memory happens to read as: "nobody wrote here" is not the same claim as
	// "this reads as empty", and only the second one is what a spec relies on.
	img.clearGoroutines()
	img.clearThreads()
	return img
}

// clearGoroutines makes the runtime.allgs slice header read as absent.
func (r *runtimeImage) clearGoroutines() {
	r.u64(r.allgs, 0)
	r.u64(r.allgs+8, 0)
	r.u64(r.allgs+16, 0)
}

func (r *runtimeImage) clearThreads() { r.u64(r.allm, 0) }

// unreadableRuntime makes every read of the goroutine table fail outright, which
// is what a stripped binary, an attach without DWARF, or a pre-init entry stop
// look like to the reader. It is expressed as a read failure rather than as
// zeroed memory because "this address reads as empty" depends on the image's
// global layout, and that differs between ELF and Mach-O — a spec that means
// "the runtime cannot be read" should not be able to pass on one platform and
// fail on the other.
func (r *runtimeImage) unreadableRuntime(unreadable bool) {
	if !unreadable {
		r.fb.failReadFrom, r.fb.failReadTo = 0, 0
		return
	}
	r.fb.failReadFrom, r.fb.failReadTo = r.allgs, r.allgs+24
}

func (r *runtimeImage) u64(addr uint64, v uint64) {
	var b [8]byte
	binary.LittleEndian.PutUint64(b[:], v)
	r.fb.seedMem(addr, b[:])
}

func (r *runtimeImage) u32(addr uint64, v uint32) {
	var b [4]byte
	binary.LittleEndian.PutUint32(b[:], v)
	r.fb.seedMem(addr, b[:])
}

func gAddr(i int) uint64 { return gStructBase + uint64(i)*gStructSpan }

// seedGoroutines plants n goroutines with goids 1..n. sliceLen is what the
// slice header advertises, which is how a test drives the reader past its scan
// ceiling without allocating that many structs. current names the goroutine
// whose stack brackets the seeded SP.
func (r *runtimeImage) seedGoroutines(n int, sliceLen uint64, current int) {
	r.current = current
	r.u64(r.allgs, gArrayBase)
	r.u64(r.allgs+8, sliceLen)
	r.u64(r.allgs+16, sliceLen) // cap

	for i := 0; i < n; i++ {
		goid := i + 1
		g := gAddr(i)
		r.u64(gArrayBase+uint64(i)*8, g)
		r.u64(g+uint64(r.off["g.goid"]), uint64(goid))
		r.u32(g+uint64(r.off["g.atomicstatus"]), gStatusRunning)
		// Every goroutine but the root is spawned by goid 1, so the current
		// goroutine's ancestor chain is short and the tree is well formed.
		if goid != 1 {
			r.u64(g+uint64(r.off["g.parentGoid"]), 1)
		}
		if goid == current {
			r.u64(g+uint64(r.off["g.stack"])+uint64(r.off["stack.lo"]), currentStackLo)
			r.u64(g+uint64(r.off["g.stack"])+uint64(r.off["stack.hi"]), currentStackHi)
			continue
		}
		// A disjoint, non-containing stack window keeps exactly one goroutine
		// current.
		lo := uint64(0x8000_0000) + uint64(i)*0x2000
		r.u64(g+uint64(r.off["g.stack"])+uint64(r.off["stack.lo"]), lo)
		r.u64(g+uint64(r.off["g.stack"])+uint64(r.off["stack.hi"]), lo+0x1000)
	}
}

// seedThreads plants a terminated runtime.allm linked list of n Ms.
func (r *runtimeImage) seedThreads(n int) {
	if n == 0 {
		r.u64(r.allm, 0)
		return
	}
	r.u64(r.allm, mStructBase)
	for i := 0; i < n; i++ {
		m := mStructBase + uint64(i)*mStructSpan
		r.u64(m+uint64(r.off["m.procid"]), uint64(1000+i))
		r.u64(m+uint64(r.off["m.id"]), uint64(i))
		// The first M runs the current goroutine, which is what makes it the
		// current thread and a required packing anchor.
		if i == 0 && r.current != 0 {
			r.u64(m+uint64(r.off["m.curg"]), gAddr(r.current-1))
		}
		next := uint64(0)
		if i+1 < n {
			next = mStructBase + uint64(i+1)*mStructSpan
		}
		r.u64(m+uint64(r.off["m.alllink"]), next)
	}
}

var _ = Describe("goroutine event production", func() {
	var (
		fb  *fakeBackend
		d   debugger.Debugger
		img *runtimeImage
	)

	BeforeEach(func() {
		bin, err := inspectFixture()
		Expect(err).NotTo(HaveOccurred())

		fb = newFakeBackend()
		d = debugger.NewWithBackend(fb, nil)
		debugger.ExportedLoadDWARF(d, bin)
		fb.seedRegs(debugger.Registers{SP: currentSP, PC: 0x1234})
		debugger.ExportedForceSuspended(d)
		img = newRuntimeImage(fb, d)
	})

	AfterEach(func() {
		_ = d.Kill()
		if !fb.stopped {
			close(fb.stopCh)
			fb.stopped = true
		}
	})

	Describe("scan clipping", func() {
		It("reports neither collection clipped on a walk that completes", func() {
			img.seedGoroutines(8, 8, 3)
			img.seedThreads(4)

			snap, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(snap.Goroutines).To(HaveLen(8))
			Expect(snap.Threads).To(HaveLen(4))
			Expect(snap.Current).To(Equal(3))
			Expect(snap.Totals).To(BeNil(),
				"a complete result must stay byte-identical to the unbounded shape")
		})

		It("flags only the goroutine scan when runtime.allgs outruns its ceiling", func() {
			// The slice claims more goroutines than the reader will walk; only
			// the ones it can reach are backed by real structs.
			img.seedGoroutines(64, debugger.ExportedMaxGoroutineScan+1, 3)
			img.seedThreads(4)

			snap, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(snap.Totals).NotTo(BeNil(),
				"a clipped scan must say so even when nothing was trimmed for size")
			Expect(snap.Totals.GoroutinesClipped).To(BeTrue())
			Expect(snap.Totals.ThreadsClipped).To(BeFalse(),
				"the thread walk finished, so calling its count approximate would be a lie")
			Expect(snap.Totals.Goroutines).To(Equal(len(snap.Goroutines)),
				"totals report what was scanned, not what the ceiling was")
		})

		It("flags the thread scan when the M list outruns its ceiling", func() {
			img.seedGoroutines(8, 8, 3)
			img.seedThreads(debugger.ExportedMaxThreadScan + 16)

			snap, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(snap.Totals).NotTo(BeNil())
			Expect(snap.Totals.ThreadsClipped).To(BeTrue())
			Expect(snap.Totals.GoroutinesClipped).To(BeFalse(),
				"the goroutine walk finished, so calling its count approximate would be a lie")
			Expect(snap.Totals.Threads).To(BeNumerically("<=", debugger.ExportedMaxThreadScan))
		})
	})

	Describe("bounding", func() {
		It("keeps a saturating snapshot inside the wire contract", func() {
			// More goroutines than the element cap admits, so the packer must
			// trim — and the event must still fit both ceilings.
			img.seedGoroutines(protocol.MaxSnapshotGoroutines+500,
				uint64(protocol.MaxSnapshotGoroutines+500), 3)
			img.seedThreads(64)

			snap, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(len(snap.Goroutines)).To(BeNumerically("<=", protocol.MaxSnapshotGoroutines))
			Expect(snap.Totals).NotTo(BeNil())
			Expect(snap.Totals.Goroutines).To(Equal(protocol.MaxSnapshotGoroutines + 500))
			Expect(snap.Totals.GoroutinesClipped).To(BeFalse(),
				"the walk completed; only the frame was trimmed")

			evt := protocol.MustEvent(protocol.EventGoroutineSnapshot, 1, snap)
			raw, err := protocol.MarshalEvent(evt)
			Expect(err).NotTo(HaveOccurred())
			Expect(len(raw)).To(BeNumerically("<=", protocol.MaxGoroutineEventBytes))

			Expect(snap.Current).To(Equal(3), "the current goroutine is a required anchor")
			Expect(goidsOf(snap.Goroutines)).To(ContainElement(3))
			Expect(goidsOf(snap.Goroutines)).To(ContainElement(1),
				"the current goroutine's whole spawn chain must survive")
		})

		It("keeps a saturating goroutine list inside the wire contract", func() {
			img.seedGoroutines(protocol.MaxSnapshotGoroutines+500,
				uint64(protocol.MaxSnapshotGoroutines+500), 3)
			img.seedThreads(4)

			payload, err := d.Goroutines()
			Expect(err).NotTo(HaveOccurred())
			Expect(len(payload.Goroutines)).To(BeNumerically("<=", protocol.MaxSnapshotGoroutines))
			Expect(payload.Totals).NotTo(BeNil())
			Expect(payload.Totals.Goroutines).To(Equal(protocol.MaxSnapshotGoroutines + 500))
			Expect(payload.Goroutines).NotTo(BeEmpty(),
				"a flat list must degrade by truncation, never to empty — an empty "+
					"threads response makes the DAP translator invent a synthetic main")

			evt := protocol.MustEvent(protocol.EventGoroutines, 1, payload)
			raw, err := protocol.MarshalEvent(evt)
			Expect(err).NotTo(HaveOccurred())
			Expect(len(raw)).To(BeNumerically("<=", protocol.MaxGoroutineEventBytes))
		})
	})

	Describe("lifecycle accounting", func() {
		It("diffs the runtime, not the trimmed frame", func() {
			// A first snapshot seeds the baseline; the second adds goroutines
			// beyond the element cap. If the deltas were computed against the
			// packed subset, every trimmed goroutine would look like it exited.
			img.seedGoroutines(protocol.MaxSnapshotGoroutines,
				uint64(protocol.MaxSnapshotGoroutines), 3)
			img.seedThreads(4)
			first, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(first.Created).To(BeEmpty(), "the first snapshot has no baseline to diff")
			Expect(first.Exited).To(BeEmpty())

			img.seedGoroutines(protocol.MaxSnapshotGoroutines+8,
				uint64(protocol.MaxSnapshotGoroutines+8), 3)
			second, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(second.Created).To(HaveLen(8),
				"exactly the eight new goroutines, regardless of what the frame carried")
			Expect(second.Exited).To(BeEmpty(),
				"trimming a goroutine off the wire must never look like it exited")
		})

		It("cannot produce deltas that overflow the contract", func() {
			// The reason the non-conforming path is unreachable in production:
			// both deltas are set differences over live sets the reader itself
			// caps, so neither can exceed the delta ceiling. If a future change
			// raises the scan without raising the ceiling, this fails here
			// rather than as a dead observer in the field.
			Expect(debugger.ExportedMaxGoroutineScan).To(BeNumerically("<=", protocol.MaxLifecycleDeltaIDs))
		})
	})

	Describe("degraded snapshots", func() {
		It("falls back to the synthetic goroutine when the runtime is unreadable", func() {
			// An unreadable runtime.allgs: the pre-init entry stop, a stripped
			// binary, or attach-without-DWARF.
			img.seedThreads(4)
			img.unreadableRuntime(true)

			snap, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			// The composed contract. The stop is still reported as the legacy
			// single synthetic goroutine rather than erroring, AND the counts
			// that stand for a runtime nobody could read are marked as the floor
			// they are: absent totals mean "complete and exact" on the wire, so
			// emitting none here would claim the runtime holds exactly one
			// goroutine and no threads. "At least one" is the honest reading
			// whether the runtime is merely not up yet or a scan failed partway.
			Expect(snap.Goroutines).To(HaveLen(1))
			Expect(snap.Goroutines[0].ID).To(Equal(1))
			Expect(snap.Goroutines[0].Status).To(Equal("waiting"))
			Expect(snap.Totals).NotTo(BeNil(),
				"a degraded read must not present itself as a complete, exact count")
			Expect(snap.Totals.Goroutines).To(Equal(1))
			Expect(snap.Totals.Threads).To(Equal(0))
			Expect(snap.Totals.GoroutinesClipped).To(BeTrue())
			Expect(snap.Totals.ThreadsClipped).To(BeTrue())
			Expect(snap.Created).To(BeEmpty(), "a degraded read invents no lifecycle events")
			Expect(snap.Exited).To(BeEmpty())
		})

		It("marks the degraded goroutine LIST as a lower bound too", func() {
			// The snapshot is not the only degraded shape on the wire: the
			// Goroutines query answers a DAP threads request from the same
			// synthetic stand-in, and absent totals there make the same claim of
			// an exact count. Both degraded paths must read as a floor, or the
			// honesty depends on which request the client happened to send.
			img.seedThreads(4)
			img.unreadableRuntime(true)

			payload, err := d.Goroutines()
			Expect(err).NotTo(HaveOccurred())
			Expect(payload.Goroutines).To(HaveLen(1))
			Expect(payload.Goroutines[0].ID).To(Equal(1))
			Expect(payload.Totals).NotTo(BeNil(),
				"a degraded list must not present itself as a complete, exact count")
			Expect(payload.Totals.Goroutines).To(Equal(1))
			Expect(payload.Totals.GoroutinesClipped).To(BeTrue())
		})

		It("does not let an unreadable runtime look like every goroutine exited", func() {
			img.seedGoroutines(8, 8, 3)
			img.seedThreads(4)
			_, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())

			// Make the goroutine table unreadable so the next snapshot degrades.
			img.unreadableRuntime(true)
			degraded, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(degraded.Exited).To(BeEmpty())

			// The baseline must be untouched, so restoring the runtime reports
			// no spurious creations either.
			img.unreadableRuntime(false)
			restored, err := d.GoroutineSnapshot()
			Expect(err).NotTo(HaveOccurred())
			Expect(restored.Created).To(BeEmpty())
			Expect(restored.Exited).To(BeEmpty())
		})
	})

})

func goidsOf(gs []protocol.Goroutine) []int {
	out := make([]int, 0, len(gs))
	for _, g := range gs {
		out = append(out, g.ID)
	}
	return out
}
```

</details>

<details>
<summary><code>export_test_seams.patch</code></summary>

```diff
diff --git a/internal/debugger/export_test.go b/internal/debugger/export_test.go
index 7cfad56..38e51dc 100644
--- a/internal/debugger/export_test.go
+++ b/internal/debugger/export_test.go
@@ -96,3 +96,66 @@ func ExportedSetBreakpointAtErr(d Debugger, addr uint64) error {
 		return err
 	})
 }
+
+// ExportedRuntimeVarAddr resolves a runtime package variable's address through
+// the loaded DWARF, so a test can plant a synthetic runtime.allgs / runtime.allm
+// in the fake backend's memory at exactly the address the reader will look at.
+func ExportedRuntimeVarAddr(d Debugger, name string) (uint64, bool) {
+	e := d.(*engine)
+	var (
+		addr  uint64
+		found bool
+	)
+	_ = e.dispatch(func() error {
+		if e.dw == nil {
+			return nil
+		}
+		addr, found = e.dw.runtimeVarAddr(name)
+		return nil
+	})
+	return addr, found
+}
+
+// ExportedGoOffsets returns the DWARF-resolved runtime struct offsets the
+// snapshot reader uses, keyed "<struct>.<field>". A test fabricating g and m
+// structs must lay them out at these offsets rather than hardcoding any, for
+// the same reason the reader does: they shift between Go versions.
+func ExportedGoOffsets(d Debugger) (map[string]int64, bool) {
+	e := d.(*engine)
+	out := make(map[string]int64)
+	ok := false
+	_ = e.dispatch(func() error {
+		l, valid := e.getGoLayout()
+		if !valid {
+			return nil
+		}
+		ok = true
+		out["g.stack"] = l.gStack
+		out["g.m"] = l.gM
+		out["g.sched"] = l.gSched
+		out["g.atomicstatus"] = l.gAtomicstatus
+		out["g.goid"] = l.gGoid
+		out["g.waitreason"] = l.gWaitreason
+		out["g.parentGoid"] = l.gParentGoid
+		out["g.gopc"] = l.gGopc
+		out["g.startpc"] = l.gStartpc
+		out["stack.lo"] = l.stackLo
+		out["stack.hi"] = l.stackHi
+		out["gobuf.pc"] = l.bufPC
+		out["m.procid"] = l.mProcid
+		out["m.curg"] = l.mCurg
+		out["m.id"] = l.mID
+		out["m.alllink"] = l.mAlllink
+		out["m.spinning"] = l.mSpinning
+		return nil
+	})
+	return out, ok
+}
+
+// ExportedMaxGoroutineScan and ExportedMaxThreadScan expose the reader's walk
+// ceilings so a test can prove the lifecycle deltas they bound cannot overflow
+// the wire contract.
+const (
+	ExportedMaxGoroutineScan = maxGoroutineScan
+	ExportedMaxThreadScan    = maxThreadScan
+)
```

</details>

<details>
<summary><code>fakebackend_hook.patch</code></summary>

```diff
diff --git a/internal/debugger/engine_test.go b/internal/debugger/engine_test.go
index 47f89c9..a002b4c 100644
--- a/internal/debugger/engine_test.go
+++ b/internal/debugger/engine_test.go
@@ -3,6 +3,7 @@ package debugger_test
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"math"
 	"syscall"
 	"testing"
@@ -34,6 +35,14 @@ type fakeBackend struct {
 	singleStepCalls  []int
 	stopProcessCalls int
 	writtenAt        map[uint64][]byte
+
+	// failReadFrom/failReadTo make ReadMemory fail for any read overlapping a
+	// half-open address range, so a test can reach the reader's "the walk
+	// stopped early" branches. The range is bounded rather than open-ended
+	// because a real image's runtime globals live above any synthetic region.
+	// Disabled while failReadTo is zero.
+	failReadFrom uint64
+	failReadTo   uint64
 }
 
 func newFakeBackend() *fakeBackend {
@@ -84,6 +93,9 @@ func (f *fakeBackend) SingleStep(tid int) error {
 }
 
 func (f *fakeBackend) ReadMemory(addr uint64, dst []byte) error {
+	if f.failReadTo != 0 && addr < f.failReadTo && addr+uint64(len(dst)) > f.failReadFrom {
+		return fmt.Errorf("fakeBackend: unmapped address 0x%x", addr)
+	}
 	for i := range dst {
 		dst[i] = f.mem[addr+uint64(i)]
 	}
```

</details>

### Verification status

`gofmt` clean. Green at #223 `0d242b9`; the composed-contract assertions were verified against a local composition of #196's then-tip, driving three mutants — degraded snapshot `PackSnapshot(..., false, false)`, degraded list `clipped = false`, and `readThreads` ceiling-only — all of which failed as intended.

**Re-run on the final base is the #196 owner's call.** Two of those mutants target #196's own code, so only its final tip proves them. The third (`readThreads`) belongs to #219/#224 now and its spec is not in this bundle.

I hold these artifacts frozen pending receipt acknowledgement.

xsachax:
## Completing the payload set — 4th artifact inlined so a 4/4 comparison is possible

Comment [5239010011](https://github.com/bingosuite/bingo/pull/223#issuecomment-5239010011) inlines **three** payloads (`goroutines_producer_test.go`, `export_test_seams.patch`, `fakebackend_hook.patch`) and lists `composed-contract-correction.patch` in the manifest **by hash only** — its content was omitted because it is already applied inside the `.go` file.

That makes a literal 4/4 inline comparison impossible against that comment alone: an owner checking four artifacts finds three blocks and one bare hash. Inlining the fourth here closes it.

**`composed-contract-correction.patch`** — 2833 bytes, `sha256 = 96b71c5d85d8a7141028cbc02511018a867fb977c658aa909e1f57cf70a653a0`

> ⚠️ **Do not apply this patch.** It is **already applied** inside the `goroutines_producer_test.go` published in `5239010011`. Applying it again will fail. It is inlined solely so the fourth hash can be verified against content rather than taken on trust.

<details>
<summary><code>composed-contract-correction.patch</code> — provenance only, already applied</summary>

```diff
diff --git a/internal/debugger/goroutines_producer_test.go b/internal/debugger/goroutines_producer_test.go
index 4ce2120..90bbe75 100644
--- a/internal/debugger/goroutines_producer_test.go
+++ b/internal/debugger/goroutines_producer_test.go
@@ -351,18 +351,45 @@ var _ = Describe("goroutine event production", func() {
 
 			snap, err := d.GoroutineSnapshot()
 			Expect(err).NotTo(HaveOccurred())
-			// The documented contract: the stop is still reported, as the legacy
-			// single synthetic goroutine, rather than erroring. What Totals says
-			// about a snapshot that carries no real runtime data is a property of
-			// the packer, covered by its own suite, and deliberately not pinned
-			// from here.
+			// The composed contract. The stop is still reported as the legacy
+			// single synthetic goroutine rather than erroring, AND the counts
+			// that stand for a runtime nobody could read are marked as the floor
+			// they are: absent totals mean "complete and exact" on the wire, so
+			// emitting none here would claim the runtime holds exactly one
+			// goroutine and no threads. "At least one" is the honest reading
+			// whether the runtime is merely not up yet or a scan failed partway.
 			Expect(snap.Goroutines).To(HaveLen(1))
 			Expect(snap.Goroutines[0].ID).To(Equal(1))
 			Expect(snap.Goroutines[0].Status).To(Equal("waiting"))
+			Expect(snap.Totals).NotTo(BeNil(),
+				"a degraded read must not present itself as a complete, exact count")
+			Expect(snap.Totals.Goroutines).To(Equal(1))
+			Expect(snap.Totals.Threads).To(Equal(0))
+			Expect(snap.Totals.GoroutinesClipped).To(BeTrue())
+			Expect(snap.Totals.ThreadsClipped).To(BeTrue())
 			Expect(snap.Created).To(BeEmpty(), "a degraded read invents no lifecycle events")
 			Expect(snap.Exited).To(BeEmpty())
 		})
 
+		It("marks the degraded goroutine LIST as a lower bound too", func() {
+			// The snapshot is not the only degraded shape on the wire: the
+			// Goroutines query answers a DAP threads request from the same
+			// synthetic stand-in, and absent totals there make the same claim of
+			// an exact count. Both degraded paths must read as a floor, or the
+			// honesty depends on which request the client happened to send.
+			img.seedThreads(4)
+			img.unreadableRuntime(true)
+
+			payload, err := d.Goroutines()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(payload.Goroutines).To(HaveLen(1))
+			Expect(payload.Goroutines[0].ID).To(Equal(1))
+			Expect(payload.Totals).NotTo(BeNil(),
+				"a degraded list must not present itself as a complete, exact count")
+			Expect(payload.Totals.Goroutines).To(Equal(1))
+			Expect(payload.Totals.GoroutinesClipped).To(BeTrue())
+		})
+
 		It("does not let an unreadable runtime look like every goroutine exited", func() {
 			img.seedGoroutines(8, 8, 3)
 			img.seedThreads(4)
```

</details>

### Complete artifact set, all four now content-verifiable

| Artifact | Bytes | SHA-256 | Inline location |
| --- | --- | --- | --- |
| `goroutines_producer_test.go` | 15585 | `c541909ff385eb9b92a89b12cf910fc150a2930f1221b47ffcee4ae9425e4a67` | comment 5239010011 |
| `export_test_seams.patch` | 2229 | `f14965a3322fdb73be9883ade62b7957da2bc7230f0fa9d3411f04f24fcd6aad` | comment 5239010011 |
| `fakebackend_hook.patch` | 1229 | `f7520369f3f84ec2fdf66e052c09c716495386b7f9bcbe428f74587b0a621c01` | comment 5239010011 |
| `composed-contract-correction.patch` | 2833 | `96b71c5d85d8a7141028cbc02511018a867fb977c658aa909e1f57cf70a653a0` | **this comment** |

Together these two comments carry the full transfer with no dependency on any session directory.

### On pinning the acknowledgement

Prefer the **four payload hashes** over a comment-body hash. Payload hashes are content-addressed; a comment hash changes if the comment is ever edited, and byte counts differ between GitHub's API and UI renderings (22,982 via API vs 22,953 as counted elsewhere) — which would look like a mismatch when nothing had actually changed.

xsachax:
## Durable transfer supplement: composed-contract correction provenance

This supplements comment `5239010011`. That comment's table lists four artifacts, but its fenced payload contains only the first three. The missing fourth artifact is reproduced in full below.

| Artifact | Bytes | SHA-256 |
| --- | ---: | --- |
| `composed-contract-correction.patch` | 2833 | `96b71c5d85d8a7141028cbc02511018a867fb977c658aa909e1f57cf70a653a0` |

**Do not apply this patch during the transfer.** Its changes are already incorporated into the `goroutines_producer_test.go` payload in comment `5239010011`; this standalone copy is retained only to make the provenance record complete and independently reconstructible.

```diff
diff --git a/internal/debugger/goroutines_producer_test.go b/internal/debugger/goroutines_producer_test.go
index 4ce2120..90bbe75 100644
--- a/internal/debugger/goroutines_producer_test.go
+++ b/internal/debugger/goroutines_producer_test.go
@@ -351,18 +351,45 @@ var _ = Describe("goroutine event production", func() {
 
 			snap, err := d.GoroutineSnapshot()
 			Expect(err).NotTo(HaveOccurred())
-			// The documented contract: the stop is still reported, as the legacy
-			// single synthetic goroutine, rather than erroring. What Totals says
-			// about a snapshot that carries no real runtime data is a property of
-			// the packer, covered by its own suite, and deliberately not pinned
-			// from here.
+			// The composed contract. The stop is still reported as the legacy
+			// single synthetic goroutine rather than erroring, AND the counts
+			// that stand for a runtime nobody could read are marked as the floor
+			// they are: absent totals mean "complete and exact" on the wire, so
+			// emitting none here would claim the runtime holds exactly one
+			// goroutine and no threads. "At least one" is the honest reading
+			// whether the runtime is merely not up yet or a scan failed partway.
 			Expect(snap.Goroutines).To(HaveLen(1))
 			Expect(snap.Goroutines[0].ID).To(Equal(1))
 			Expect(snap.Goroutines[0].Status).To(Equal("waiting"))
+			Expect(snap.Totals).NotTo(BeNil(),
+				"a degraded read must not present itself as a complete, exact count")
+			Expect(snap.Totals.Goroutines).To(Equal(1))
+			Expect(snap.Totals.Threads).To(Equal(0))
+			Expect(snap.Totals.GoroutinesClipped).To(BeTrue())
+			Expect(snap.Totals.ThreadsClipped).To(BeTrue())
 			Expect(snap.Created).To(BeEmpty(), "a degraded read invents no lifecycle events")
 			Expect(snap.Exited).To(BeEmpty())
 		})
 
+		It("marks the degraded goroutine LIST as a lower bound too", func() {
+			// The snapshot is not the only degraded shape on the wire: the
+			// Goroutines query answers a DAP threads request from the same
+			// synthetic stand-in, and absent totals there make the same claim of
+			// an exact count. Both degraded paths must read as a floor, or the
+			// honesty depends on which request the client happened to send.
+			img.seedThreads(4)
+			img.unreadableRuntime(true)
+
+			payload, err := d.Goroutines()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(payload.Goroutines).To(HaveLen(1))
+			Expect(payload.Goroutines[0].ID).To(Equal(1))
+			Expect(payload.Totals).NotTo(BeNil(),
+				"a degraded list must not present itself as a complete, exact count")
+			Expect(payload.Totals.Goroutines).To(Equal(1))
+			Expect(payload.Totals.GoroutinesClipped).To(BeTrue())
+		})
+
 		It("does not let an unreadable runtime look like every goroutine exited", func() {
 			img.seedGoroutines(8, 8, 3)
 			img.seedThreads(4)
```

#196 remains frozen until #224 has merged and the coordinator supplies an exact replay parent. This comment does not authorize implementation or publication.

Issue #205, opened 2026-08-09
[debugger/linux] Process-global wait4 lets sessions steal ptrace stops and wedge each other


xsachax:
**Describe the bug**
Every Linux debugger session independently calls process-global `wait4(-1, __WALL)`. Linux permits any thread in a thread group to wait on children and ptrace tracees owned by its sibling threads, so one bingo session can consume another session's stop or exit status.

This violates session isolation in the shared-server architecture. Confirmed outcomes include:

- A suspended Kill blocks indefinitely in `reapAfterKill`, because its `ECHILD` termination condition remains false while any peer session has a live tracee.
- A normal session waiter consumes a peer's exit, leaving the rightful session without `EventProcessExited`.
- Launch can wedge when another waiter consumes its initial exec stop before the targeted `Wait4(pid)` receives it.
- Foreign stops can poison a backend's `lastStopTID`, causing resume errors while the owning tracee remains parked.

Priority: **P1 / high**. The failure permanently wedges engine and hub loops and can prevent clean server shutdown.

**To Reproduce**
1. On linux/amd64, create two debugger sessions in the same bingo server process.
2. Leave session B alive while session A is suspended.
3. Kill session A.
4. Observe A block in `reapAfterKill`'s `Wait4(-1, __WALL)` instead of returning after all of A's own threads are reaped.
5. In the companion normal-operation proof, make B exit while A has an active waiter; observe A consume B's status and B fail to report its exit, or observe B's Launch block because its initial stop was consumed.

Native proof: https://github.com/bingosuite/bingo/actions/runs/31327906883

The run includes green single-session and sequential-session controls. Deterministic peer-child/peer-ptrace-stop cases reproduce the wedge; the racing peer-exit case produced six lost exits and one Launch wedge across seven iterations. A direct ownership proof recorded three peer statuses consumed by session A while the rightful waiter received `ECHILD`.

**Expected behavior**
A session must receive only statuses for its own process/thread set. Kill must terminate based on that session's owned TIDs rather than process-global `ECHILD`, and one session's launch, stop, exit, or teardown must not affect another session.

**Screenshots**
N/A.

**Desktop (please complete the following information):**
 - OS: Linux
 - Architecture: amd64
 - Version: `main` at `0509b53bfc44da889e152ee20f56ce3663882c65`; native proof used Go 1.25.5

**Additional context**
Root cause and code locations:

[...]

PR #225, opened 2026-08-09
fix(debugger): safely detach Linux targets

### LINKED ISSUE #204: [debugger/linux] Running attach teardown leaves breakpoint traps and destroys the target
**Describe the bug**
Tearing down a linux/amd64 attach session while its tracee is running attempts to restore software breakpoints and detach without first putting the tracee in a ptrace stop. Linux rejects both `PTRACE_POKEDATA` and `PTRACE_DETACH` with `ESRCH` in that state, but bingo discards both errors and reports Kill success.

The original instruction remains replaced by `INT3` (`0xcc`). When the target later executes it, the target either parks briefly in an unobserved ptrace stop or is implicitly detached as the tracer thread exits; it then receives a native fatal `SIGTRAP`. The durable outcome is that bingo silently corrupts and destroys a process it explicitly does not own.

This affects explicit Kill and ordinary lifecycle teardown: last-client disconnect, server shutdown, and idle shutdown all call `Debugger.Kill`.

Priority: **P1 / high**.

**To Reproduce**
1. On linux/amd64, start a Go target independently. Have it wait on a gate before entering a known noinline function.
2. Attach bingo to the process and set a software breakpoint inside the gated function.
3. Continue so the tracee is running but has not executed the breakpoint.
4. Kill/close the attach session while it is running.
5. Observe that Kill returns nil.
6. Compare target memory with the ELF: memory still contains `0xcc` where the ELF contains the original byte.
7. Release the gate. Observe fatal `SIGTRAP: trace trap` at `breakpoint+1` rather than normal target completion.

Native proof: https://github.com/bingosuite/bingo/actions/runs/31326501407

Controls in the same proof show suspended detach succeeds and running teardown without an armed breakpoint does not corrupt the target. Direct syscall probes show running `PTRACE_POKEDATA` and `PTRACE_DETACH` both return `ESRCH`; stopped operations succeed. Reproduced under race and non-race builds. Repeated harness setup later wedged in Attach, but the completed iterations include byte-level trap evidence, target stderr at the exact trap PC, and silent Kill success.

**Expected behavior**
Before an attached process is released, bingo must quiesce it, restore every software breakpoint, rewind the PC when parked on a breakpoint, and detach successfully. The target must remain unmodified, untraced, and making progress. Any failure to restore or detach must be surfaced rather than converted to success.

**Screenshots**
N/A.

 - OS: Linux
 - Architecture: amd64
 - Version: `main` at `0509b53bfc44da889e152ee20f56ce3663882c65`; native proof used Go 1.25.5

**Additional context**
Root cause and code locations:

- `internal/debugger/engine.go:227-251`: `Kill` calls breakpoint cleanup and process teardown while `stateRunning`.
- `internal/debugger/breakpoint.go:117-123`: `clearAll` discards every restore error.
- `internal/debugger/backend_linux_amd64.go:343-355`: restore uses `PTRACE_POKEDATA`, which requires a ptrace stop.
- `internal/debugger/backend_linux_amd64.go:185-220`: the attached branch discards `PTRACE_DETACH` errors and returns nil.
- `internal/debugger/process.go:53-66`: process state becomes non-live before teardown succeeds, preventing a meaningful retry.
- `internal/hub/hub.go:240-244,543`: normal session removal/shutdown reaches this path.

Two independent adversarial reviews confirmed the defect and corrected the scope: the report should lead with the leaked trap and silent target destruction, not claim that `TracerPid` remains permanently nonzero. Kernel implicit detach on tracer-thread exit makes persistent tracing timing-dependent, while the trap leak is deterministic.

A safe fix must coordinate with the existing one-shot `waitLoop` rather than starting a second waiter. When running, the engine should consume any already-buffered stop or request a bounded stop and read that result directly from `stopCh`; it must handle intermediate stops, rewind a breakpoint PC, restore traps with checked errors, and detach with checked errors. Teardown state and event closure must remain coherent even when cleanup fails.

Permanent linux E2E coverage should attach, arm a future breakpoint, continue, tear down, then assert: original text bytes restored, `TracerPid == 0`, target heartbeat still advances, and final status is not SIGTRAP. A suspended-at-breakpoint variant should protect the PC-rewind path.

xsachax:
> **Draft — do not merge.** This is a from-scratch restack of issue #204 on current `main`, not a replay of the stale implementation.

## Exact restack

- Base: `48ce9bfa2c926087b4c5e67a965275db254c3d8d`
- Published head: `2d45c9fba6a6cbd8bd69209b551b91e2ee3a3e38`
- Canonical branch: `xsachax-fix-linux-attach-teardown`
- Initial force-with-lease replaced exactly `a2a723e88bc116dcf122079cb844cc59823d7eb4`
- Base retargeted from the obsolete stack to `main`

## Defect

A running Linux attach teardown restored breakpoint bytes and detached while its TIDs were not ptrace-stopped. Linux returned `ESRCH` for both `PTRACE_POKEDATA` and `PTRACE_DETACH`; bingo discarded both errors, cleared `process.live`, and reported success. The foreign process retained `INT3` and later died with native `SIGTRAP`.

## Current-main design

1. **One wait owner:** every engine waiter starts through tracked `startWait`; attached teardown cancels and joins that waiter before synchronously consuming the existing `linuxWaitOwner`. Production Linux `Wait4` remains broker-only.
2. **Whole-group attach:** every pre-existing TID is `PTRACE_SEIZE`d with `TRACECLONE|TRACEEXEC|TRACEEXIT`, registered by exact broker generation, interrupted, and drained. Stable stop→scan passes catch pre-option and late clones; auto-attached clone `EPERM` is accepted only when `TracerPid` matches the dedicated tracer thread TID.
3. **Provenance-preserving quiesce:** parked stops, held step-owner anchors, pending signals, `PTRACE_EVENT_STOP`, group stops, terminal statuses, and queued step completions retain their exact TID/reason. The active `stepTID` is captured before folding so a completed instruction is never rewound and executed twice.
4. **Repair→restore→detach:** delayed bingo breakpoint PCs are repaired exactly once, a verified stopped write anchor is selected, breakpoint restoration is all-or-nothing and checked, then non-leaders detach before the leader. Exec skips stale-byte restoration.
5. **Signal fidelity:** current/backlog/delayed signal batches stay per-TID. Genuine delivery signals are passed only from their matching stop; deferred signals and job-control group stops are requeued with exact-TID `tgkill`; breakpoint/interrupt/Pause signals are suppressed only for detach. Normal resume semantics are unchanged.
6. **Retryable partial failure:** exact detach errors propagate. Successfully released TIDs are re-seized and re-quiesced after a partial failure; `process.live`, the engine, and tracer remain owned. Commands other than retrying `Kill` are rejected until cleanup completes.
7. **Lifecycle retention:** a focused hub test proves a retained partial attach would otherwise be dropped. Hub/server completion is withheld while `ErrAttachedDetachIncomplete` is retryable; failed startup cleanup retries in the background so the original Attach error still reaches the client. `Server.Shutdown` can return `ErrShutdownIncomplete` while `Done` stays open. A second CLI signal is the explicit operator escape hatch.

## Commit series

| Commit | Purpose |
| --- | --- |
| `2caa949` | Core per-TID seize/quiesce/restore/detach redesign and unit coverage |
| `19326df` | Retain incomplete cleanup through hub/server shutdown |
| `47526fc` | Native attached-teardown E2E |
| `5aceeaf` | Document new invariants |
| `8b15a5d`, `54dc6c6` | Preserve quiesce-time and running job-control group stops |
| `ccd1652` | Keep failed-start cleanup asynchronous to the hub loop |
| `a454206`, `738f3a5` | Simplify result handling and satisfy lint/error contracts |
| `38388cc` | Preserve queued single-step identity during quiesce |
| `92bac0e` | Add exact per-TID timeout diagnostics |
| `0b82f89` | Keep the native proof's `Cmd.Wait` from competing with the broker |
| `07c61c1` | Include per-TID state on both quiesce timeout exits |
| `2d45c9f` | Deduplicate the attach harness without reintroducing a wait consumer |

## Negative control on exact current main

Isolated proof commit `752b9e1075b6626b2eaeb65332968747d347f715` had sole parent `48ce9bfa…` and no production fix. Native hosted-Linux run [31829177451](https://github.com/bingosuite/bingo/actions/runs/31829177451) recorded:

- suspended control: `POKEDATA err=<nil>`, `DETACH err=<nil>`, `TracerPid … after=0`
- running defect: `POKEDATA=no such process`, `DETACH=no such process`, `TracerPid … after=2852`
- explicit verdict: `U4 SYSCALL PREMISE CONFIRMED`

The imported old high-level proof harness itself timed out in current-main `Debugger.Attach`, so it is not counted as evidence. The raw syscall paired control above is the durable fail-before proof. The temporary proof branch was deleted.

### Local

- `go tool goimports -l .` and `git diff --check`
- host + E2E integration vet; full Linux/amd64 `-tags e2e` vet
- Linux/amd64 `go build ./...`
- Linux/amd64 `golangci-lint run ./...`: **0 issues**
- `go test -tags bingonative -race ./...`: full repository green
- new Linux detach/broker/engine tests: 25 race-stress repetitions green
- hub/server/DAP: 10 complete Linux race rounds green
- native Apple Silicon `just e2e-darwin` on `2d45c9fb…`: **24 passed, 0 failed, 1 opt-in diagnostic skipped**; churn 200, basic 25, kill 20, attach 3, fullstack 15, hygiene task send refs `1→1`

### Native hosted Linux / GitHub

[Debugger E2E run 31848828552](https://github.com/bingosuite/bingo/actions/runs/31848828552) is fully green:

- Linux amd64 ptrace job: **pass** (17m14s), including attach, kill, exit, signals, five churn rounds, and overlap
- Full-stack Linux: **pass**
- Darwin compile jobs/full-stack controls: **pass**
- The attach step runs the existing repeated attach spec plus running-armed, suspended-at-breakpoint, and no-breakpoint teardown controls. All four pass.

The first native run exposed a test-harness bug rather than a product failure: its early `exec.Cmd.Wait` was a second wait4 consumer and stole the leader's interrupt stop. Per-TID diagnostics showed the kernel in tracing-stop with an empty broker queue; delaying `Cmd.Wait` until after detach made the exact-head rerun green and now permanently pins the single-consumer invariant.

## Independent reviews

- Rubber-duck design review identified `PTRACE_EVENT_STOP`, queue-drain, signal-delivery, terminal-generation, and shutdown-retention hazards before implementation; all were incorporated.
- Independent implementation reviews found auto-clone `EPERM`, ordinary wait-error ownership loss, exec stale-byte restore, interrupt `SIGTRAP` injection, park-time signal publication, job-control group stops, failed-start hub blocking, and quiesced step-owner misclassification. Each received a focused regression and exact-head re-review.
- Final exact-head review: **No findings**.

## Complete GitHub matrix

| Workflow | Run | Result |
| --- | --- | --- |
| Go (build, imports, vet, lint, unit + race) | [31848828525](https://github.com/bingosuite/bingo/actions/runs/31848828525) | Pass |
| Debugger E2E | [31848828552](https://github.com/bingosuite/bingo/actions/runs/31848828552) | Pass |
| CodeQL | [31848828480](https://github.com/bingosuite/bingo/actions/runs/31848828480) | Pass |
| VS Code packaging | [31848828550](https://github.com/bingosuite/bingo/actions/runs/31848828550) | Pass |
| Darwin policy tests | [31848828482](https://github.com/bingosuite/bingo/actions/runs/31848828482) | Pass |
| Darwin exact-head verification status | [31849120714](https://github.com/bingosuite/bingo/actions/runs/31849120714) | Pass |
| SonarCloud | PR check | Pass |

## Residual risks / explicit limits

- Attached teardown is deadline-bounded per attempt but shutdown intentionally retains ownership and retries rather than falsely completing; operators may force process exit with a second signal.
- The detach transaction stops every owned TID but does not make ordinary debugging an atomic stop-the-world debugger outside teardown.
- Wire protocol and management API shapes are unchanged.

Fixes #204.


xsachax:
The shared `lastStopTID` ownership prerequisite is now issue #231 / draft base PR #232 from `main`.

Final stack order should be **#232 -> #202 -> #225**. The stale #225 head's broad tracee mutex happens to serialize the plain field, but #204 does not subsume the defect: launched Kill and running Set/Clear remain reachable outside attach teardown. When restacking, preserve #225's tracee-state mutex/selection semantics but consume #232's synchronized field ownership rather than carrying an independent race fix.

Base evidence:
- unfixed proof `31338374195` and natural #202 workload `31329090064`
- fixed head `6ffa54b`, race run `31338972574` / job `93309465646`
- atomic-removal mutation `31339235577` / job `93310150239`
- native debugger matrix `31338972568`, all jobs green

#205 remains the separate process-global wait-status broker boundary.

PR #229, opened 2026-08-09
fix(wsmon): escape carriage returns

### LINKED ISSUE #228: [wsmon] Bare carriage returns can overwrite rendered status text
**Describe the bug**

`cmd/wsmon` renders tracee-controlled output, panic text, and error messages as single terminal lines. Its shared `oneLine` helper makes LF visible as the two printable characters `\n`, but leaves a bare carriage return (`\r`, byte `0x0d`) intact. A terminal interprets that byte by returning the cursor to column zero, so the remainder of the message can overwrite the `last: ` prefix or earlier rendered content.

A second one-line path bypasses `oneLine`: under `-once`, a rejected `CmdGoroutineSnapshot` is retained as advisory context and later interpolated into the final stderr error. A carriage return in that `ErrorPayload.Message` can likewise overwrite the leading `error: ` text.

**Code locations**

- `cmd/wsmon/main.go`, `oneLine`: only `"\n"` is escaped before trimming and truncation.
- `cmd/wsmon/main.go`, `describeEvent`: panic, output, and ordinary error events pass arbitrary payload text through `oneLine` before being assigned to `lastEvent`.
- `cmd/wsmon/main.go`, `monitor.render`: `lastEvent` is printed as `fmt.Printf("last: %s\n\n", ...)`.
- `cmd/wsmon/main.go`, `rejectionContext`: snapshot rejection text is interpolated into a one-line error without passing through `oneLine`.
- `cmd/wsmon/main_test.go`, `TestOneLine/embedded_crlf`: the current test deliberately characterizes the raw-CR behavior left out of PR #210.

**Impact**

Any debug target that writes progress-style output containing carriage returns, panics with such text, or causes an error containing one can corrupt wsmon's display. Labels and earlier status text can be hidden or replaced, making telemetry misleading and harder to diagnose. This affects terminal presentation; it does not alter the debug target or server state.

**To Reproduce**

1. Check out the head of PR #210 (`xsachax-fix-wsmon-utf8-truncation`, commit `8019b58`).
2. Temporarily add this focused test in `cmd/wsmon/main_test.go`:

   ```go
   func TestOneLineBareCarriageReturnRepro(t *testing.T) {
       got := oneLine("before\rafter")
       want := `before\rafter`
       if got != want {
           t.Fatalf("oneLine result = %q, want %q", got, want)
       }
   }
   ```

3. Run:

   ```sh
   go test -tags bingonative ./cmd/wsmon -run TestOneLineBareCarriageReturnRepro -count=1
   ```

4. The failure reports that the actual string contains `\r`, while the expected string contains the printable characters `\\r`.
5. To observe the terminal effect directly, run:

   ```sh
   printf 'last: before\rafter\n'
   ```

   The text after the carriage return starts at column zero and overwrites the beginning of the line.

**Observed output**

The helper returns `"before\rafter"` with a raw `0x0d`. A normal terminal visually renders the direct reproducer as `after before`, replacing `last:` with `after`. Longer suffixes can overwrite more of the preceding content. The `-once` rejection-context path has the same behavior for the leading `error: ` line.

**Expected behavior**

`oneLine` should contain no raw CR or LF. It should preserve the exact control sequence visibly: CR as `\r`, LF as `\n`, CRLF as `\r\n`, and repeated controls as repeated escapes. Escaping should happen before existing whitespace trimming and before PR #210's rune-budget truncation, matching current LF behavior: leading/trailing line controls remain visible, ordinary surrounding whitespace is trimmed, and the normalized output remains at most 120 runes without introducing invalid UTF-8. Snapshot rejection context should use the same helper rather than duplicate or bypass these semantics.

**Screenshots**

N/A; the exact byte sequence and terminal command above reproduce the issue deterministically.

- OS: macOS and Linux terminals
- Architecture: darwin/arm64 and linux/amd64
- Version: PR #210 head (`8019b58`)

**Additional context**

PR #210 fixes UTF-8 truncation and explicitly keeps newline normalization byte-for-byte unchanged, including the surviving CR in its `embedded crlf` characterization test. This report is intentionally a separate stacked follow-up. Canonicalizing CRLF to one `\n` would be lossy; independently escaping CR and LF preserves whether the target emitted CR, LF, CRLF, LFCR, or repeated controls. As with the existing LF behavior, truncation may split the two-character visible escape at the rune-budget boundary; that remains valid UTF-8 and preserves PR #210's established budget semantics.

xsachax:
## Summary

`cmd/wsmon` now renders every raw carriage return as the visible characters `\r`, matching its existing LF escaping and preventing tracee-controlled text from moving the terminal cursor back to column zero. CR and LF are escaped independently, so CRLF, LFCR, and repeated controls retain their exact sequence.

The `-once` snapshot-rejection context now uses the same `oneLine` helper; it previously bypassed normalization and could overwrite the leading `error: ` text even after the main helper was fixed.

Fixes #228

## Stack

`#210 → this PR`

This draft targets `xsachax-fix-wsmon-utf8-truncation` (#210), not `main`. Review only commit `26d6700`; merge after #210.

## Semantics

- Escape CR as `\r` and LF as `\n` before trimming.
- Preserve CRLF as `\r\n`, LFCR as `\n\r`, and repeated controls one-for-one rather than canonicalizing them lossily.
- Keep leading/trailing line controls visible while continuing to trim surrounding spaces and tabs.
- Apply PR #210's 120-rune budget after normalization. As already accepted for LF, truncation may split a two-character escape but never splits a UTF-8 code point.
- Preserve PR #210's valid-input and invalid-input behavior: valid UTF-8 remains valid, and already-invalid bytes pass through unchanged.

## Tests

The table covers bare CR, LF, CRLF, LFCR, repeated and trailing controls, trim ordering, exact/over normalized budgets, cuts inside CR/LF escapes, ASCII/CJK/emoji/combining sequences, and invalid UTF-8. Every case asserts the 120-rune ceiling, valid-input preservation, and absence of raw CR/LF. An end-to-end `-once` test pins the snapshot-rejection bypass.

- `go test -tags bingonative -race -count=1 ./cmd/wsmon/...`
- `go build -tags bingonative ./...`

Independently reviewed for root cause and normalization semantics; the review identified the snapshot-rejection bypass included in this patch.

### DIFF
--- a/cmd/wsmon/main.go
+++ b/cmd/wsmon/main.go
@@ -208,7 +208,7 @@ func rejectionContext(msg string) string {
 	if msg == "" {
 		return ""
 	}
-	return fmt.Sprintf(" (a snapshot request was rejected meanwhile: %s)", msg)
+	return fmt.Sprintf(" (a snapshot request was rejected meanwhile: %s)", oneLine(msg))
 }
 
 func (m *monitor) applyEvent(evt protocol.Event) (bool, bool) {
@@ -535,8 +535,10 @@ const (
 	oneLineKeepRunes = oneLineMaxRunes - len("...")
 )
 
+var oneLineEscaper = strings.NewReplacer("\r", `\r`, "\n", `\n`)
+
 func oneLine(s string) string {
-	s = strings.TrimSpace(strings.ReplaceAll(s, "\n", `\n`))
+	s = strings.TrimSpace(oneLineEscaper.Replace(s))
 	if s == "" {
 		return "-"
 	}
--- a/cmd/wsmon/main_test.go
+++ b/cmd/wsmon/main_test.go
@@ -135,6 +135,26 @@ func TestRunOnceDeadlineReportsObservedRejection(t *testing.T) {
 	}
 }
 
+func TestRunOnceEscapesRejectionContext(t *testing.T) {
+	events := make(chan protocol.Event, 1)
+	events <- protocol.MustEvent(protocol.EventError, 2, protocol.ErrorPayload{
+		Command: protocol.CmdGoroutineSnapshot,
+		Message: "before\rafter\nlast",
+	})
+	close(events)
+
+	m := &monitor{clients: -1}
+	err := m.run(events, true, time.Minute, func() {})
+	if err == nil {
+		t.Fatal("run(-once) returned nil despite never receiving a snapshot")
+	}
+	if got := err.Error(); !strings.Contains(got, `before\rafter\nlast`) {
+		t.Fatalf("run error = %q, want escaped rejection context", got)
+	} else if strings.ContainsAny(got, "\r\n") {
+		t.Fatalf("run error = %q, want no raw line controls", got)
+	}
+}
+
 // The deadline also bounds a silent server, with no rejection to report.
 func TestRunOnceDeadlineWithoutRejection(t *testing.T) {
 	events := make(chan protocol.Event)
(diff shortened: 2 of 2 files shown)


xsachax:

PR #232, opened 2026-08-09
fix(debugger): synchronize linux stop TID access

### LINKED ISSUE #231: [debugger/linux] lastStopTID races between waitLoop and running memory operations
**Describe the bug**

At `origin/main` commit `0509b53bfc44da889e152ee20f56ce3663882c65`, `linuxBackend.recordStop` writes the plain `lastStopTID` field from the one-shot `waitLoop` goroutine while engine-loop commands can read it through `traceTID` before the stop crosses `stopCh`. The normal stopped-state handoff is ordered, but running operations have no happens-before edge.

The confirmed production path is:

`engine.waitLoop -> linuxBackend.Wait -> recordStop` racing with
`engine.Kill -> breakpointTable.clearAll -> WriteMemory -> traceTID`.

The other reachable readers are `WriteMemory` from running `SetBreakpoint`/`ClearBreakpoint` and the `ReadMemory` ptrace fallback from running `SetBreakpoint`. `ContinueProcess` also reads `traceTID`, but its current call sites are stopped-state ordered: it runs before a new wait loop starts or after the prior wait result is received.

This is a production Go data race and therefore undefined behavior. Independently of the race detector, a torn/stale target selection can make a TID-less ptrace memory operation target a thread whose stop state changed and fail with `ESRCH`; `Kill` currently discards breakpoint restore failures, while Set/Clear surface them.

**To Reproduce**

1. On native linux/amd64, check out proof commit `587722926a340e0a62ba7fcf7226041d208b7fad`. Its only changes from `0509b53` are focused tests and a proof workflow; production code is unchanged.
2. Run the stopped-state control:
   `go test -race -count=1 -v ./internal/debugger -run '^TestLinuxBackendStoppedMemoryStopTIDOrdered$'`
3. Observe that ordered `WriteMemory`, ptrace-fallback `ReadMemory`, and `ContinueProcess` complete without a race report.
4. Run the production-path probes:
   `go test -race -count=1 -v ./internal/debugger -run '^TestLinuxBackendRunning(Kill|Memory)StopTIDRace$'`
5. Observe race reports for running Kill, direct WriteMemory, and ptrace-fallback ReadMemory.

Native proof run `31338374195` passed the stopped-state control and deterministically failed all three running reader variants under `-race`. The workflow intentionally inverted that expected failure, so the overall run is green.

A separate real-ptrace workload independently captured the natural production stack in run `31329090064`:

```text
Write: linuxBackend.recordStop
       linuxBackend.drainParked
       linuxBackend.Wait
       engine.waitLoop

Read:  linuxBackend.traceTID
       linuxBackend.WriteMemory
       breakpointTable.clear
       breakpointTable.clearAll
       engine.Kill.func1
       engine.loop
```

The focused current-main proof captured the same Kill reader, plus:

```text
traceTID -> WriteMemory -> TestLinuxBackendRunningMemoryStopTIDRace/write-memory
traceTID -> ReadMemory  -> TestLinuxBackendRunningMemoryStopTIDRace/read-memory-fallback
```

All race against `recordStop -> stopTIDRaceBackend.Wait`; the embedded backend uses the production `recordStop`, `traceTID`, memory methods, engine actor, Kill path, and breakpoint table.

**Expected behavior**

`lastStopTID` must have explicit cross-goroutine ownership. A reader should obtain one defined snapshot while `Wait` publishes a newly stopped TID; the existing ptrace operation must then either run against that snapshot or return its normal syscall error. Stopped-state operations must retain their channel-ordered behavior.

The permanent regression should be mutation-discriminating under `-race`: it must fail when the field/accessors are reverted to a plain `int`, pass when synchronized, cover Kill plus both memory readers, and retain the stopped-state negative control.

**Screenshots**

N/A; the failure is captured by native race-detector stacks and workflow logs.

 - OS: Ubuntu 24.04, kernel `6.17.0-1020-azure` (GitHub-hosted native runner)
 - Architecture: linux/amd64
 - Version: production code from `0509b53bfc44da889e152ee20f56ce3663882c65`; proof head `587722926a340e0a62ba7fcf7226041d208b7fad`; Go 1.25.5

**Additional context**

Code locations on current main:

- `internal/debugger/backend_linux_amd64.go:94-100` — plain `lastStopTID` field
- `internal/debugger/backend_linux_amd64.go:319-354` — ptrace-fallback ReadMemory and WriteMemory readers
- `internal/debugger/backend_linux_amd64.go:429-587` — Wait runs on the wait-loop goroutine and records returned stops
- `internal/debugger/backend_linux_amd64.go:592-607` — `setPID`, `traceTID`, and `recordStop`
- `internal/debugger/engine.go:221-250` — running Kill clears breakpoints before process teardown
- `internal/debugger/breakpoint.go:117-123` — `clearAll` reaches WriteMemory

Dependency/duplication audit:

- #199 / draft PR #202 owns sibling-stop delivery corruption. PR #202 incidentally found this race in run `31329090064` and currently carries an `atomic.Int64` fix in commit `04f7f9a`, but issue #199 does not own the cross-goroutine access defect and the race also exists on unmodified main outside its parked-stop queue.
- #204 / draft PR #225 owns unsafe running attach teardown. Its current stale-base mutex happens to serialize this field, but its declared final stack requires restacking on #202's atomic ownership. Quiescing attached teardown does not cover launched Kill or running Set/Clear, so #204 does not subsume this issue.
- #205 owns process-global `wait4(-1, WALL)` cross-session status theft. Synchronizing this field does not solve foreign wait ownership.

The smallest base fix should land independently beneath #202 and #225; those branches can then drop their incidental copies when restacked. Two independent adversarial reviews confirmed the access map, the stopped-state happens-before control, and that #204 cannot be the sole owner. One review cautioned—correctly—that synchronization defines access, not global stop-state semantics; this fix must not be described as making arbitrary ptrace memory operations safe against a running tracee.

xsachax:
Closes #231

## Summary

- make Linux `lastStopTID` an `atomic.Int64` because `Wait` writes it from the one-shot `waitLoop` goroutine while running engine commands can read it through `traceTID`
- cover the exact running **launched** `Kill -> clearAll -> WriteMemory` path plus direct `WriteMemory` and ptrace-fallback `ReadMemory` readers
- retain stopped-state controls for WriteMemory, ReadMemory fallback, and ContinueProcess
- keep the native race proof process-safe: invalid negative backend IDs, `proc.pid=0`, an explicit non-nil launched marker, and unconditional idempotent cleanup
- select one stable parent regression suite in native Linux CI
- document that the atomic is a defined snapshot, not tracee quiescence or a stop-state guarantee

## Root cause and scope

Normal suspended operations are ordered by the `waitLoop -> stopCh -> engine loop` handoff. Running commands bypass that handoff:

```text
waitLoop: recordStop -> lastStopTID.Store
engine:   Kill -> clearAll -> WriteMemory -> traceTID.Load
```

Other reachable readers are WriteMemory from running SetBreakpoint/ClearBreakpoint and ReadMemory's PTRACE_PEEKDATA fallback from running SetBreakpoint. ContinueProcess reads the same field, but current call sites are stopped-state ordered. The dedicated tracer thread serializes ptrace syscalls; it does not synchronize the Go field read before the handoff.

## Test safety and non-vacuity

The regression never names a real process:

- backend PID and published stop TIDs are strictly negative, so `process_vm_readv` is skipped and ptrace cannot target a live process
- the engine process keeps `cmd != nil` so #204 still treats it as launched, but `pid=0` makes `process.kill` return before `killProcess` or any signal
- `sync.Once` cleanup always stops and drains the fake waiter, including `Fatalf` paths
- the stopped control records a distinct negative TID before its ordered reads
- `TestLinuxBackendStopTIDRaceRegressions` is the stable CI entry point; helper renames cannot silently omit a case

Two exact-head adversarial reviews covered signal/process safety, fallback non-vacuity, Kill path fidelity, cleanup, mutation sensitivity, and CI selection. The final review found no issues.

## Native evidence

- `31338374195`, head `5877229`: unmodified production code reported races for running Kill, WriteMemory, and ReadMemory fallback while the stopped control stayed clean
- `31329090064`: real ptrace workload captured `recordStop -> drainParked -> Wait -> waitLoop` racing `traceTID -> WriteMemory -> clearAll -> Kill -> engine.loop`
- independent same-address runs `31337440091` and `31337440952`: `Wait -> recordStop` raced running `SetBreakpoint -> breakpointTable.set -> WriteMemory -> traceTID`; `31337441789` reproduced corruption without a race report
- final safe fixed head `8807dfeaae67c2d03a244fbdef63eb964026423d`: Go run `31349219793`, debugger matrix `31349219798`, and VS Code run `31349219777` all passed
- exact safe atomic-removal mutation `c78a59b666f593884a00e0083b9cad05def8fb7f`: run `31349410269`, job `93337291325`, independently caught running Kill, WriteMemory, and ReadMemory fallback; the stopped control passed

The address-reservation guard proposed for the same-address proof is complementary, not a substitute: it prevents one caller, while this PR defines publication for every reachable running `traceTID` reader.

## Stack ownership and ancestry

Published linear stack:

- #232: `8807dfeaae67c2d03a244fbdef63eb964026423d`
- #184: `45dfa4f79ab15e3f2217e0765f20127df7d32e84`
- #202: `b9771e10145d676abdfe7dd1954b0abb11d1b945`
- #225: `a3f451370628d13d781ea38243898799562c66ba`

#202 owns sibling-stop delivery; its duplicate atomic commit remains dropped. #225 owns safe attached teardown; attached Kill gets a stronger quiesced-stop guarantee, while launched Kill and running Set/Clear retain this atomic prerequisite. #205 remains separate process-global wait ownership work.

## Validation

- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c ./internal/debugger`
- `go tool goimports` / `git diff --check`
- all PR checks green and head mergeable at `8807dfeaae67c2d03a244fbdef63eb964026423d`

The PR remains draft for stack-owner review, not for missing technical evidence.


xsachax:
## Native A/B complete

- Fixed head `6ffa54b`: run `31338972574`, job `93309465646` passed the exact permanent `-race` selector for running Kill, WriteMemory, ReadMemory fallback, and stopped Write/Read/Continue controls.
- Atomic-removal mutation `4f03d8c`: run `31339235577`, job `93310150239` kept the final tests unchanged, reverted only the production field/accessors, passed the stopped control, and produced race reports in all three running paths.
- Native debugger run `31338972568` passed all four jobs. Every PR check is green and the head is mergeable.

The proof-only mutation branch has been deleted; the run and logs remain available.

xsachax:
## Stack hardening update

Head `f37659002448e70c17105ee18d8516eeb27eb46c` makes the Kill race regression explicitly launched (`process.cmd != nil`). This keeps the inherited test on launched teardown after #204 gives attached Kill an ordered quiesce-before-clear transaction.

Native Go run `31343912384`, job `93322176089` passed the exact focused `-race` selector, including the hardened Kill path. Two unrelated matrix jobs are being rerun: Linux churn hit its documented 180s target watchdog at iteration 197/200, and VS Code's Electron setup timed out downloading over the network. Neither failure touched this test or diff.

xsachax:
## Hardened launched-path mutation gate

The post-#204 hardening is now validated on both sides:

- fixed frozen head `f37659002448e70c17105ee18d8516eeb27eb46c`: native Go run `31343912384`, job `93322176089`, exact focused `-race` selector green
- atomic-removal mutation `4595ba57cf1bc25b0ef15b4ee6d58eff89d784dd`: run `31344374799`, job `93323467414`; stopped-state control green, and the explicitly launched Kill plus both running memory readers all produced race reports

Relative to `f376590`, the mutation changed only the production stop-TID field/accessors back to the original plain `int` and added the proof workflow. The proof branch is being deleted; Actions evidence remains.

All #232 checks are green after rerunning two unrelated failures: the pre-existing 180s churn target-watchdog cliff and an Electron download network timeout.

xsachax:
Independent same-address proof strengthens the running-memory coverage:

- `31337440091` and `31337440952`: native race stacks show `Wait -> recordStop` racing `SetBreakpoint -> breakpointTable.set -> WriteMemory -> traceTID`
- `31337441789`: reproduces breakpoint corruption without a race report

The proposed address-reservation guard for that proof is complementary, not a substitute: it prevents one same-address call, while this PR defines publication for every reachable running `traceTID` reader. Head remains `f37659002448e70c17105ee18d8516eeb27eb46c`; all checks are green and the PR is mergeable. The post-change adversarial review found no incomplete accessor conversion, atomic-copy hazard, selector gap, or stop-state overclaim.

xsachax:
Process-safety blocker resolved at final head `8807dfeaae67c2d03a244fbdef63eb964026423d`. Two exact-head adversarial reviews approve; the final review found no issue. Fixed validation: Go `31349219793`, debugger matrix `31349219798`, VS Code `31349219777`. Exact atomic-removal mutation `c78a59b` run `31349410269` caught running Kill, WriteMemory, and ReadMemory fallback while the stopped control remained clean. Final green stack: #184 `45dfa4f79ab15e3f2217e0765f20127df7d32e84` → #202 `b9771e10145d676abdfe7dd1954b0abb11d1b945` → #225 `a3f451370628d13d781ea38243898799562c66ba`.

xsachax:
Final #225 owner cross-validation confirms the boundary documented here: attached running Kill quiesces owned TIDs and consumes `waitPending` before breakpoint `WriteMemory`; that path gets a stronger stop-state guarantee from #225. Launched running Kill and running Set/Clear memory operations remain outside that transaction and depend on this PR’s atomic publication. #232 is the sole owner of general `lastStopTID` synchronization.

PR #243, opened 2026-08-12
fix(debugger): preserve cleared breakpoints after owner death

xsachax:
## P1 regression

Main `83e0be812f32c476fb03e4cd9346641728889c28` can resurrect a successfully cleared software breakpoint when the thread single-stepping off it exits. During the step-off window the entry is retained in `steppingOverBP`; `ClearBreakpoint` disables that entry because the original instruction is already restored. Both owner-death completion paths nevertheless reinstalled it unconditionally:

- the cross-platform `StopBreakpoint` death backstop
- the Linux `StopStepThreadExited` reconciliation boundary

That violates the API invariant that a successful clear is final and can make a cleared breakpoint fire again. A sibling can also have an already-queued same-address trap stop from before the clear, so simply skipping reinstall is insufficient: the delayed stop must be recognized without mistaking a genuine live trap for stale state.

## Fix

- Reinstall retained entries on owner death only while they remain enabled.
- Always complete step bookkeeping and release the Linux owner-death gate, including the cleared-entry path.
- Retain the restored instruction bytes for cleared in-flight entries so a delayed same-address stop is byte-verified, PC-rewound, and silently resumed.
- Preserve enabled-entry reinstall-before-release behavior and all asynchronous halt semantics from #202.
- Add focused fake-backend and engine-over-stepQueue regressions, and document the final clearing invariant in `AGENTS.md`.

## Provenance

This branch is exactly one commit on `83e0be812f32c476fb03e4cd9346641728889c28`. Its patch is identical to reviewed commit `4699d76ef3bce48e1283280a62fa9dca43007fb6`, which was not included when #202 merged from an intermediate head. Stable patch ID: `f1ffc3f0762e13595ee942c36b0645012644bdd0`. No old #202 stack commits are present.

## Validation

- `gofmt` and `goimports` clean
- Three focused owner-death/clear regressions under `-race`
- `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 golangci-lint run` (0 issues)
- `just e2e-darwin` (22/22 passed)

Native Linux E2E and mutation execution require the Linux CI runners; the existing fresh-PR gates remain unchanged.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -367,10 +367,13 @@ invalidates matching `lastBP` / `lastBPTID`; the live PC was rewound when the
 trap stop arrived, so the next resume can continue directly from the original
 instruction. During an in-flight step-off the entry is intentionally absent
 from the table and its original bytes are already restored: `ClearBreakpoint`
-matches the retained entry by ID, marks it disabled, and succeeds. Completion
-then skips `reinstall` but still performs the saved `bpResumeAction`. Failed
-restoration leaves the entry enabled and all pending state intact. `clearAll`
-applies the same invalidation/cancellation rules for Kill.
+matches the retained entry by ID, records its restored bytes, marks it disabled,
+and succeeds. Completion then skips `reinstall` but still performs the saved
+`bpResumeAction`. Recording the bytes keeps an already-queued same-address
+sibling hit classifiable after owner death: it is rewound and resumed silently,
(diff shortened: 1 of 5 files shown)


xsachax:
Superseded by merged PR #242, which carries the same focused owner-death breakpoint fix on current main with full green CI and Darwin verification.

PR #263, opened 2026-09-17
fix(debugger): retire Darwin per-session Mach namespaces

### LINKED ISSUE #233: [debugger/darwin] Each debug session permanently leaks ~6 Mach port names; unbounded namespace growth in long-lived servers
**Priority: P2** (no repo priority labels exist, so recording it here.)

> **Frozen queue — #234 lands first in full. No arm of this issue is authorized before it.**
>
> **#234** — *[debugger/darwin] Attach never saves the target's `EXC_MASK_BREAKPOINT` handler, so detach permanently wedges or kills the victim process* — must be **implemented, exact-head verified, and merged** before any implementation work on this issue is authorized.
>
> The dependency is hard, not stylistic: the fix for this issue must release `excPort`, and destroying `excPort` while a **live attached target** still references it as its installed exception handler kills that target. Until #234 restores the victim's prior handler on detach, there is no safe point at which this issue's teardown can free `excPort` on a live, non-owned process.
>
> **The gate is not merely "#234 is open with a plan"**, and it is **not per-arm**. No implementation arm of this issue — including the launch/Kill arm — is authorized until #234's **full restoration + latch + distinct wake + acknowledged join** is implemented, exact-head verified, and merged. There is deliberately **no carve-out** allowing the launch/Kill teardown to land ahead of it: the two arms share `acquirePorts`, the same port state, and the same waiter lifecycle, so partial ordering here is how a subtle teardown race gets introduced.
>
> Once #234 is merged, **this issue owns a deliberately narrow scope**: post-COMPLETE namespace destruction and census. It does not own the waiter lifecycle, which #234 establishes. Teardown here must additionally **consult the outstanding-detach registry** before destroying any right, so a session with an outstanding detach is never torn down out from under.
>
> As of this writing #234 is unapproved and under independent review, so nothing here is authorized. Any teardown landed before it would reintroduce, in a harder-to-diagnose form, exactly the use-after-free and victim-kill hazards this ordering exists to prevent.
>
> This ordering does not affect the claim below. The measured namespace growth is from **launch/Kill and launch/natural-exit** sessions and stands on its own as evidence; only the *fix* is sequenced.

## Describe the bug

On `darwin/arm64`, every bingo debug session permanently leaks **~6 Mach port names** into the *debugger/server* process's IPC namespace. The names are never released, because the Darwin backend has **no teardown path at all** — `acquirePorts` allocates per session and nothing ever unwinds it.

Growth is unbounded and monotonic in a long-lived server process, which is exactly the model the VS Code companion's connect-or-start + `-idle-timeout` design encourages, and it is compounded by the ordinary `Restart` lifecycle (hub-level Restart discards the old `Debugger` and constructs a fresh one, so each restart is another full session's worth of leaked names).

Per session, measured by port-name identity (not just counts), present in **30/30 cycles with zero variance**:

| leaked name | right class | per session | source |
| --- | --- | --- | --- |
| `portSet` | **PORT-SET** | 1.000 | `mach_darwin_arm64.h:321` |
| `excPort` | send+recv | 1.000 | `mach_darwin_arm64.h:300`/`:302` |
| `notePort` | send+recv | 1.000 | `mach_darwin_arm64.h:308`/`:310` |
| `ctrlPort` | send+recv | 1.000 | `mach_darwin_arm64.h:316`/`:318` |
| `taskPort` | **DEAD-NAME** | 1.000 | `backend_darwin_arm64.go:262` |
| `threadPorts[…]` | **DEAD-NAME** | 1.000 (floor) | `backend_darwin_arm64.go:765` |

**~6/session is a floor, not a constant.** The dead-name component scales with the number of *distinct faulting threads* per session (`adoptExcThreadPort` retains a right per thread; each becomes a dead name when the tracee dies), so breakpoint-heavy or thread-churning sessions leak more.

### Scope of the claim (deliberately conservative)

What is proven here is **unbounded namespace growth in the shared long-lived server** — nothing more.

- This issue does **not** claim a measured per-task namespace ceiling.
- It does **not** claim `KERN_UREFS_OVERFLOW`.
- It does **not** claim an observed `Wait` wedge.

Those were the *symptoms* of #100 and are not re-demonstrated by this experiment. The defect asserted here is the resource leak itself.

## Not a duplicate of #100

Closed **#100** ("Mach exception messages leak task & thread send rights every stop") fixed **per-stop descriptor uref growth** — the user-reference count on already-known names climbing on every exception. That fix is intact and unaffected:

- Send **urefs grew +2.9987/session**, i.e. exactly **one uref per newly leaked name** — no uref inflation on existing names.
- The existing `hygiene` gate (`declarePortHygieneSpec`, `debugger_e2e_darwin_arm64_test.go:87`) re-run **unmodified passes**: `base=1 after=1` over 40 breakpoint stops.

This issue is **per-session namespace *names***, which that gate is blind to by construction: it samples `DarwinTaskPortSendRefs` — the uref count on a *single* session's task port — and therefore cannot observe cross-session growth in the number of names. Orthogonal axis, orthogonal defect.

Also distinct from **#204**, which is linux-only (ptrace attach teardown).

Related to but distinct from **#234**: #234 is a *correctness* defect in the victim process (the target's exception handler is never restored on detach). This issue is a *resource* defect in the debugger process (namespace names are never released). They share the `bingo_setup_exception_ports` call site, which is why they must be sequenced — see the stack ordering note at the top — but neither subsumes the other.

## To Reproduce

Run repeated launch→run→Kill sessions in one long-lived process and census the debugger's own Mach namespace between cycles.

`lsmp -p <pid>` was **unusable** on the test host: it calls `task_for_pid` on its target and fails for *every* process including the invoking shell, with no passwordless sudo available. Substituted the in-process equivalent — `mach_port_names(mach_task_self(), …)` + `mach_port_type` + `mach_port_get_refs` — which needs no privileges and yields lsmp's exact right-class breakdown for the debugger process itself.

```sh
git rev-parse HEAD   # 0509b53bfc44da889e152ee20f56ce3663882c65

# Build + codesign exactly as the justfile e2e-darwin recipe does
env CGO_ENABLED=1 go test -tags 'e2e bingonative' -race -c -o ./build/bingo-e2e.test ./test/integration
codesign --sign - --entitlements entitlements.plist --force ./build/bingo-e2e.test
codesign -d --entitlements - ./build/bingo-e2e.test

# Temporary untracked census helpers (preserved under the artifact path below):
#   internal/proofcensus/census_darwin_arm64.go          cgo Mach census + release primitives
#   internal/debugger/zz_proof_ports_darwin_arm64.go     exports the backend's port NAMES
#   test/integration/zzz_portproof_darwin_arm64_test.go  the specs

BINGO_PROOF_OUT=$OUT BINGO_PROOF_SESSIONS=60 \
  ./build/bingo-e2e.test -test.v -test.timeout 20m \
  -ginkgo.label-filter=portproof-debug -ginkgo.v

BINGO_PROOF_OUT=$OUT BINGO_PROOF_SESSIONS=60 \
  ./build/bingo-e2e.test -test.v -test.timeout 20m \
  -ginkgo.label-filter=portproof-control -ginkgo.v

# Untouched upstream gate, for contrast (passes)
./build/bingo-e2e.test -test.v -test.timeout 10m -ginkgo.label-filter=hygiene -ginkgo.v
```

## Measured results

Baseline SHA: **`0509b53bfc44da889e152ee20f56ce3663882c65`** (verified `== origin/main` at time of measurement).

### Two independent 60-session debug runs

| metric | run 1 (n=60) | run 2 (n=60) | R² |
| --- | --- | --- | --- |
| **total names / session** | **+5.9957** | **+6.0131** | 0.9999 |
| receive | +2.9921 | +2.9945 | 0.9998 |
| send | +3.0055 | +3.0186 | 0.9998 |
| **port sets** | **+1.0000** | **+1.0000** | **1.0000** |
| **dead names** | **+1.9981** | **+2.0000** | 0.9999 |

**Predicted from source = 6/session** (4 persistent from `bingo_setup_exception_ports` + ≥2 dead from `taskPort` and retained thread rights). **Measured 5.996 / 6.013** — agreement within 0.3%. Per-cycle total delta had mean 6.0000, mode 6.

The **+1.0000 port-set/session at R²=1.0000** is uniquely attributable: `mach_darwin_arm64.h:321` is the **only** `MACH_PORT_RIGHT_PORT_SET` allocation in the entire repository, so no other code path can produce that slope.

### Matched control — 60 executions, no debugger

Same target binary, same long-lived process, comparable wall-clock (30.8s vs 38.2s), plain `os/exec` launches:

| metric | slope | R² |
| --- | --- | --- |
| total names | **+0.0066** | 0.049 (noise) |
| **port sets** | **0** | — |
| **dead names** | **0** | — |

A second 30-exec control at matched duration was **exactly flat** (22 → 22 names, per-cycle delta min=0 max=0). Go/Ginkgo/`-race` runtime drift is **not** a confound.

### No delayed reclamation observed

`runtime.GC()` ×3 plus a 15 s settle after the 60-session run:

| probe | total | port sets | dead names |
| --- | --- | --- | --- |
| last cycle | 390 | 61 | 122 |
| after-gc | 391 | 61 | 122 |
| after-gc + 5 s | 391 | 61 | 122 |
| after-gc + 15 s | 392 | 61 | 122 |

Port sets and dead names pinned exactly; the total drifts *up*, never down. **No delayed reclamation was observed.**

### Independent cross-checks

- After each `Kill` + termination, a per-name `mach_port_type` query (which does **not** walk the namespace, so it independently validates the census) returned `KERN_SUCCESS` for all 6 ports in 30/30 cycles. A released name returns `KERN_INVALID_NAME`.
- A survival pass re-probing every name ever added, after GC + settle: 39/230 were transient runtime churn, and **all 180 bingo-attributed names survived**.
- Hand-releasing two finished sessions' ports reclaimed exactly 6 namespace slots (216 → 210) — you cannot release what you do not own, confirming the rights were genuinely still held and that nothing in bingo had freed them.

## Caveat on scope of the quantitative proof

The allocation is shared: **`acquirePorts` is used by both the launch and the attach paths**, and the source shows the same resources lack teardown in either case. However, the quantitative proof above covers **launch/Kill and launch/natural-exit only**.

**Live-attach exception-handler restoration is tracked separately as #234**, and safe cleanup of `excPort` on a *live* target depends on it: `bingo_setup_exception_ports` calls `task_set_exception_ports(…)` but never saves or restores the previous handler, so releasing the receive right while a live, non-owned target still has it installed is not obviously safe. The detached-target hang/crash consequence is **not claimed as proven here** — see #234 for that. **#234 must be implemented, exact-head verified, and merged — not merely designed — before any teardown work on this issue begins.** Note that this gate is not limited to the attach path: per the frozen-queue policy at the top, it blocks the launch/Kill arm as well.

## Root cause / code locations

1. **`acquirePorts` — `internal/debugger/backend_darwin_arm64.go:242`.** Runs once per session. Acquires `task_for_pid`, stores `b.taskPort` / `b.taskOK` under `taskMu` (**:261-264**), then calls `bingo_setup_exception_ports` and sets `b.portsOK` (**:266-269**).
2. **`bingo_setup_exception_ports` — `internal/debugger/mach_darwin_arm64.h:292`.** Allocates **4 names**: `excPort` (`:300` RECEIVE, `:302` MAKE_SEND), `notePort` (`:308`/`:310`), `ctrlPort` (`:316`/`:318`), `portSet` (`:321`). It also installs a task exception handler (`:304`) and requests a dead-name notification (`:312`). **There is no unwind function anywhere** — no `bingo_teardown_*`, and every early `return kr` inside it abandons whatever it already allocated.
3. **`threadPorts` adoption — `backend_darwin_arm64.go:765`** (`adoptExcThreadPort`). Retains a send right per faulting thread; each becomes a dead name when the tracee dies.
4. **Engine shutdown lacks a Darwin close/join — `internal/debugger/engine.go:515-523`.** The deferred teardown closes `done`/`events`, then calls `closeTracer()` behind an `interface{ closeTracer() }` type assertion. That is a **linux-only** hook; `darwinBackend` does not implement it, so the assertion silently fails and **Darwin runs no teardown whatsoever**.
5. **Neither kill path releases anything — `backend_darwin_arm64.go:334` (`killProcess`).** The *attached* branch (`!db.launched`, **:339**) flushes replies and resumes threads; the *launched* branch flushes, resumes, SIGKILLs and reaps. Neither releases `portSet`/`excPort`/`notePort`/`ctrlPort`/`taskPort`/`threadPorts`.
6. **No release primitive exists in the codebase for these.** `grep` over `internal/` finds **zero** `mach_port_mod_refs`, **zero** `mach_port_destroy`, and **zero** `runtime.SetFinalizer`. The only primitive used is `mach_port_deallocate`, and all 11 call sites are per-stop / per-scan uref balancing from the #100 work — none is session teardown.
7. **Existing hygiene blind spot — `test/integration/debugger_e2e_darwin_arm64_test.go:87.** As described above, samples one session's task-port urefs; structurally cannot see cross-session name growth.

### Failure-path leaks (must be covered by the fix)

- **Partial setup unwind.** Each of the ~9 early `return kr` points inside `bingo_setup_exception_ports` leaks every name allocated before it. `portsOK` stays false, so even a future teardown keyed on `portsOK` would skip them.
- **`taskPort` acquired before setup failure.** `acquirePorts` sets `b.taskPort`/`b.taskOK` at **:261-264** *before* calling setup at **:266**. If setup fails, `acquirePorts` returns an error with the task send right already stored and never released.

## Expected behavior

A finished session releases every Mach right and name it acquired. Namespace size in a long-lived server returns to baseline after each session, so port sets and dead names do not accumulate.

## Fix requirements (design, not code — implementation is blocked pending coordinator approval)

1. **Explicit Darwin backend teardown.** Introduce a real teardown entry point for the backend and invoke it from engine shutdown, rather than relying on the linux-only `closeTracer()` assertion.
2. **Shutdown latch plus a wake sentinel distinct from Pause** — **owned by #234**, consumed here. `ctrlPort` is currently used only for the Pause wake (`BINGO_MSG_PAUSE`). Teardown needs (a) a **latch** recording that shutdown has begun, so the state is durable and idempotent rather than inferred from a one-shot message, and (b) its **own wake sentinel**, so a shutdown wake can never be misread as a manual stop. This issue does not re-litigate that design; it depends on it having landed.
3. **Acknowledged waiter-join before releasing any receive right or port set** — **owned by #234**, consumed here. `Wait` blocks in `mach_msg` on the port set. Destroying a receive right or the set out from under a blocked receiver is a use-after-free class hazard. Signalling the wake is **not sufficient** — the waiter must be observed to **acknowledge** the wake and **confirm it has returned** (a join, not a best-effort nudge) before any right is released. A wake that races the waiter's next blocking `mach_msg` leaves the receiver blocked on a set that is about to be destroyed. This issue's destruction step runs strictly **after** that join has completed.
4. **Flush or discard deferred replies correctly.** The backend stashes un-acked exception replies per thread (invariant: every resume path flushes, or `Wait` hangs). Teardown must resolve every pending stashed reply before tearing down.
5. **Cancel the dead-name notification** requested against `notePort` (`mach_darwin_arm64.h:312`) before releasing it.
6. **Release full uref counts, based on the actual right type.** Measured on XNU 12377: `taskPort` carried `deadname_urefs=2`, so a single release call left it in the namespace. Teardown must query the real right type and drop **all** urefs, not call a primitive once.
7. **Destroy both send and receive rights.** A `RECEIVE`+`SEND` name (`excPort`/`notePort`/`ctrlPort`) requires **two** operations in either order — measured: `mach_port_deallocate` alone leaves the name alive holding `recv`; `mach_port_mod_refs(RECEIVE,-1)` alone leaves it alive as a `DEAD-NAME`. Neither single call suffices.
8. **Keep the #100 hygiene gate green.** The per-stop uref balancing must not regress.
9. **Destroy nothing until #234 is merged, and consult the outstanding-detach registry first.** #234 restores the target's prior `EXC_MASK_BREAKPOINT` handler on detach; without it, destroying `excPort` while a live attached target still references it kills that target. An accepted *design* is not enough — the restoration must be implemented, exact-head verified, and merged. This blocks **every** arm of this issue, launch/Kill included; there is no per-arm carve-out. Once merged, this issue's teardown runs only **post-COMPLETE** and must consult the **outstanding-detach registry** before releasing any right, so a session with an outstanding detach is never destroyed out from under.

## Regression test requirements

- A **normal-polarity native census** spec (fail-before / pass-after) that asserts namespace return-to-baseline, rather than an inverted assertion.
- **≥20 launch cycles**, and eventually **≥10 attach/detach cycles**.
- **Gate primarily on port-set and dead-name counts returning to baseline** — those are the zero-noise signals (control produced exactly 0 of each; port sets tracked at R²=1.0000). Total-name count is noisier due to benign runtime churn.
- **Census self-check / mutation teeth**: the census must be able to detect a deliberately introduced leak, so the test cannot silently pass because the measurement broke.
- **Attach forward-progress negative control**: prove the attached target is still alive and making progress after detach, so a "clean" teardown that actually broke the target cannot pass.
- **Note:** the **attach arm is required before this issue is closed** and is *not* currently proven. **No arm — launch/Kill included — may be implemented before #234 is implemented, exact-head verified, and merged.** The earlier carve-out permitting the launch/Kill arm to land first has been withdrawn: both arms share `acquirePorts`, the same port state, and the same waiter lifecycle, so splitting them is how a subtle teardown race gets introduced.

## Desktop

- **OS:** macOS 26.5.2 (25F84), Darwin 25.5.0, `xnu-12377.121.10~1/RELEASE_ARM64_T6000`
- **Architecture:** arm64 (Apple Silicon, T6000)
- **Go:** go1.26.1 darwin/arm64
- **Build:** `-tags 'e2e bingonative' -race`, adhoc-codesigned with `entitlements.plist`; `com.apple.security.cs.debugger` verified embedded
- **Commit:** `0509b53bfc44da889e152ee20f56ce3663882c65`
- **Test binary sha256:** `81cac1f7c3c09f46f13f12b7f8aef341fc0f3e2cb5b08f324689c4c2ef1ee327`

## Additional context

Full raw evidence — per-cycle CSVs by right class, per-name identity attribution, post-kill kernel probes, survival pass, release-primitive ownership matrix with raw `kern_return_t`, regression analysis, and complete run logs — is preserved locally at:

```
/Users/sacha/.copilot/session-state/c7a5ebe4-9012-4599-bbc7-d135093dd247/files/mach-port-proof/
  ├── SUMMARY.md, environment.txt, commands.sh          # run 1
  ├── raw/, parsed/, logs/, code/
  └── round2/
      ├── REPORT.md, environment.txt, commands.sh
      ├── raw/{surviving_names,post_kill_probe,survival,debug_sessions,
      │        natural_exit_sessions,control_sessions}.csv
      ├── raw/{attribution,ownership_experiment}.txt
      ├── parsed/analysis.txt, parsed/*_deltas.csv
      ├── logs/{identity,natexit,control,debug,hygiene}.log
      └── code/                                          # census helpers + analysis script
```

The investigation was strictly read-only: all helpers were untracked and removed afterwards, and the worktree was verified clean at `0509b53` with no commits, pushes, or branch changes.

An additional adversarial check worth recording: a **natural-exit** variant (tracee runs to completion, no `Kill`) leaks at **+5.9824 names/session**, with **+1.0000 port sets** and **+2.0000 dead names** per session — so the leak is a property of the **session lifecycle**, not of the `Kill` path specifically.

### LINKED ISSUE #234: [debugger/darwin] Attach never saves the target's EXC_MASK_BREAKPOINT handler, so detach permanently wedges or kills the victim process
**Describe the bug**

On darwin/arm64, `attach` installs bingo's task-level `EXC_MASK_BREAKPOINT` exception port with `task_set_exception_ports`, which **discards** the tuple it displaces. Detach (`Kill` of an attached process) then restores nothing and destroys nothing.

The victim process is left permanently pointing at a port bingo still owns but no longer services. Every subsequent breakpoint exception in that process — `runtime.Breakpoint()`, a `brk #0`, any debugger/crash handler it installed itself — is delivered to bingo's orphaned port and never answered. The target never runs again.

This is the *attach* path specifically. On launch the tracee's task dies with it, so the dangling port has no victim.

Severity: this corrupts a process bingo does not own and was explicitly asked only to observe, and it survives the session. `bingo.log` shows `from=exited to=idle` / `session removed` **before** the victim freezes — bingo believes it has no debuggee while still holding the port that wedges one.

Verified read-only against `origin/main` = `0509b53bfc44da889e152ee20f56ce3663882c65`. No production source was modified.

**To Reproduce**

1. Start a target that installs its own `EXC_MASK_BREAKPOINT` handler via `task_swap_exception_ports` (so the displaced tuple is recorded) and waits on a gate before executing `brk #0`.
2. Snapshot `task_get_exception_ports(EXC_MASK_BREAKPOINT)` from a third entitled inspector process — this is the BEFORE tuple.
3. `bingo -addr …`, then attach to the target's pid, hold, then `Kill` (detach). Keep the bingo process alive.
4. Snapshot again — the AFTER tuple.
5. Open the gate so the target executes `brk #0`.

Observed, 12/12 across 3 reps × {detach-while-suspended, detach-while-running} × {C target with own handler, Go target with none}:

```
=== SNAPSHOT label=BEFORE-attach ===
  [1] mask=0x00000040 (EXC_MASK_BREAKPOINT) port=0x00001d03 type=SEND urefs=1 behavior=EXCEPTION_DEFAULT(0x1) flavor=5
=== SNAPSHOT label=AFTER-detach ===
  [1] mask=0x00000040 (EXC_MASK_BREAKPOINT) port=0x00000c03 type=SEND urefs=1 behavior=EXCEPTION_DEFAULT(0x1) flavor=5
```

The port name changes and never changes back. The target's own self-view confirms it lost ownership: `type=0x30000` (SEND|RECEIVE, `is_own=1`) before attach becomes `type=0x10000` (SEND only, `is_own=0`) after detach. Result: `progressed=no`, `own_handler_hits=0`, target hung forever.

Ownership of the surviving port was confirmed directly by arming `MACH_NOTIFY_DEAD_NAME` on the post-detach port and `SIGKILL`ing bingo: `msgh_id=72` arrives and the victim's slot becomes `port=0xffffffff type=DEAD`. The receive right belongs to the bingo process and outlives the session.

**Two distinct failure modes, split deterministically by detach timing (3/3 each)**

- **Detach while suspended** → victim threads `suspend_count=0,0`. The exception sits queued-unread on bingo's orphaned port; the thread is blocked in-kernel awaiting a reply that never comes.
- **Detach while running** → victim threads `suspend_count=1,1`. bingo's orphaned `waitLoop` *dequeues* the exception, runs stop-the-world (`thread_suspend` on every thread), stashes the reply and drops it. A "detached" debugger leaves the victim Mach-suspended.

Killing bingo does not rescue the target — with the port dead the exception is undeliverable.

**Expected behavior**

Detach restores the target to exactly the state it was in before attach: the prior `EXC_MASK_BREAKPOINT` tuple reinstalled, its own handler working, execution continuing. Detaching from a process must never be able to hang or kill it.

**Root cause**

`internal/debugger/mach_darwin_arm64.h:304` — `bingo_setup_exception_ports`:

```c
kr = task_set_exception_ports(task, EXC_MASK_BREAKPOINT, *exc_port,
        EXCEPTION_DEFAULT, THREAD_STATE_NONE);
```

`task_set_exception_ports` discards the displaced tuple, so restoring it later is not merely omitted — it is impossible. `task_swap_exception_ports` is the call that returns it.

Supporting sites:
- `internal/debugger/backend_darwin_arm64.go:242-271` — `acquirePorts`, shared by launch and attach.
- `internal/debugger/backend_darwin_arm64.go:334-345` — the detach branch of `killProcess` does only `flushAllReplies()` + `bingo_resume_all_threads(task)`; no restore, no teardown.
- No `mach_port_deallocate` / `mach_port_mod_refs` for `excPort` / `portSet` / `notePort` / `ctrlPort` anywhere, so the receive right survives for the life of the bingo process.
- `internal/debugger/engine.go:221-251` — `Kill` sets `stateExited` and injects a synthetic `StopExited` without waking or joining the in-flight `waitLoop`, which is how the orphan waiter arises.

**Required fix ordering**

The destroy-vs-restore constraint is measured; the rest of the sequence is source-inferred and flagged inline below. Three matched experiments, 3/3 deterministic each, settle the destroy-vs-restore question. "Destroying `excPort`" was simulated by `SIGKILL` of bingo, which destroys its receive right.

| variant | teardown order tested | progressed | own-handler hits | target alive |
| --- | --- | --- | --- | --- |
| `ord1` | destroy, never restore | no | 0 | no |
| `ord2` | **restore, then** destroy | **yes** | **1** | **yes** |
| `ord3` | exception already pending → restore → destroy | no | 0 | no |

- `ord1`: the victim's slot becomes `port=0xffffffff type=DEAD` and the process dies on `brk`. **`excPort` must not be destroyed while a live attached target still references it.**
- `ord2`: after restore the slot returns to its original name and **survives bingo's death unchanged**; the target's own handler fires and it exits cleanly. **Restore-before-destroy is correct and sufficient — the port does not need to be leaked "for safety".**
- `ord3`: restoring after the exception was already delivered does not rescue the trapped thread. **The pending-reply flush is an independent obligation that must complete before destruction.**

**What is measured vs inferred.** Measured (3/3 each): steps 4 and 7 and their relative order, plus the independence of the reply flush. **Source-inferred, NOT measured:** step 1 (shutdown sentinel), step 2 (wake + join), step 5 (notification cancellation), and specifically **the detach-time drain of queued-unread exceptions in step 3**. In every experiment above the `brk` was executed *after* detach, so no exception was queued at detach time; the need to drain at detach is read off the receive-loop source, not observed. A fix should carry its own test for those steps.

Resulting order:

```
SAVE at acquire (task_swap_exception_ports)  →  … session …  →  DETACH:
  1. post a distinct SHUTDOWN sentinel — NOT the Pause/ctrl sentinel   [inferred]
  2. wake + JOIN the blocked Wait; no receiver may run past this point   [inferred]
  3. FLUSH pending replies, and drain/reply to queued-unread exceptions   [drain: inferred]
  4. RESTORE the saved prior tuple(s), including NULL   [MEASURED]
  5. CANCEL the dead-name notification   [inferred]
  6. resume threads
  7. DESTROY excPort / notePort / ctrlPort / portSet   [MEASURED: must follow 4]
```

Notes on the individual steps:

- **Step 1 must not reuse `ctrlPort`.** `mach_darwin_arm64.h:517` maps `BINGO_CTRL_MSG_ID` unconditionally to `BINGO_MSG_PAUSE`, and `backend_darwin_arm64.go:827-839` handles that by calling `stopTheWorld()` — `thread_suspend` on every thread of the victim. Waking a blocked `Wait` with `StopProcess()` during detach would make the waiter's last act be suspending the process it is detaching from, which is exactly the running-detach freeze above. A teardown wake needs a new msgh_id mapped to a terminal class that returns immediately with no `stopTheWorld`, no `reap` and no `StopEvent`. It should also latch an atomic `tearingDown` flag before posting, since `StopProcess` treats `MACH_SEND_TIMED_OUT` as benign success (`backend_darwin_arm64.go:611-620`) and could otherwise drop the wake.
- **Step 4 must replay an array, including NULL entries.** `task_swap_exception_ports` returns *n* tuples. A returned entry whose `ports[i] == MACH_PORT_NULL` is **the normal "target had no prior breakpoint handler" case, and it MUST be restored explicitly** — `task_set_exception_ports(task, masks[i], MACH_PORT_NULL, behaviors[i], flavors[i])`, which clears the slot. **Skipping null entries is a bug**: it leaves bingo's own `excPort` installed and the victim stays wedged exactly as described above.

  This is not an edge case. A stock Go target has no custom breakpoint handler at all, and the null-prior shape is directly observed: my C target's own `task_swap_exception_ports` returned `displaced 1 prior entries` / `prior[0] mask=0x40 port=0x0 behavior=0x0 flavor=0` — i.e. "no prior handler" is reported as a **present entry with a null port**, not as `count == 0`. Null-port entries are ordinary exception-table contents (the same target's table also carries `[0] mask=0x13be port=0x0 type=0x0`).

  Also handle `count == 0` by **explicitly clearing** the mask: `task_set_exception_ports(task, EXC_MASK_BREAKPOINT, MACH_PORT_NULL, EXCEPTION_DEFAULT, THREAD_STATE_NONE)`. Never treat "nothing to restore" as "do nothing".

  The only thing to skip for a null entry is the **deallocation** — there is no right to release. For non-null entries the returned send rights must be deallocated after replay, or the fix trades a wedged victim for a right leak. Restore should be gated on `!db.launched`, mirroring the existing detach branch.
- **Step 5**: `mach_port_request_notification` at `mach_darwin_arm64.h:312` must be cancelled (re-armed with `MACH_PORT_NULL`) before `notePort` dies, and the `prev_not` send-once right it returns is currently captured and dropped on the floor — it must be restored or deallocated.

**Planned decomposition**

Deliberately split so the victim-facing fix is small and reviewable on its own, and the risky lifecycle work stacks on top.

*PR1 — save/restore only (this issue).* `task_swap_exception_ports` at attach and persist the returned tuples; at detach: flush pending replies → restore the prior tuple **including the null/no-prior-handler case** → resume threads. Deallocate the non-null send rights handed back by the swap. **No port destruction and no waiter join** in this PR — `ord2` shows restore alone fully un-wedges the victim, so PR1 is sufficient to close this bug and deliberately leaves the (pre-existing) port leak to PR2.

*PR2 — lifecycle cleanup, stacked (#233).* Distinct shutdown sentinel; close `done` **before** the waiter join; perform the join **outside** `Kill`/the engine dispatch, since `Kill` runs on the engine loop thread and a `waitLoop` blocked sending to a full `stopCh` would deadlock against a join taken from that same thread; then cancel the dead-name notification and destroy the rights, with a timeout fallback of **restore-but-leak** (keeping the victim correct is strictly more important than reclaiming bingo's own port names).

The ordering constraint is why the stack runs in this direction: PR2 must not destroy `excPort` on a live attached target before PR1's restore is in place.

**Contract this establishes**

**Attach must leave the target running.** Attaching to a process bingo does not own, and later detaching from it, must be observationally transparent to that process: its prior exception-port tuple is reinstated, its own handlers keep working, and it continues executing. Detach must never be able to hang, suspend, or kill a process bingo merely observed.

**Desktop**

- OS: macOS 26.5.2, SIP enabled
- Architecture: arm64 (Apple Silicon)
- Version: `origin/main` @ `0509b53bfc44da889e152ee20f56ce3663882c65`; Go 1.26.1, Apple clang 21.0.0

**Additional context**

*Adversarial validation.* Four falsification challenges were run to try to break this result:

- **Causal repair** — a third process externally reinstalls the saved tuple after detach and changes nothing else. The target goes from permanently wedged to `progressed=yes own_handler_hits=1`, 3/3. Restoring the tuple is both necessary and sufficient for the victim.
- **LLDB differential** — attach → `process status` → `detach` against the same target leaves the slot **byte-identical** (`0x0c03 → 0x0c03`), the target keeps `type=0x30000 is_own=1`, its handler fires and it runs to completion, 3/3. This is not "just how macOS behaves"; it is specific to bingo.
- **Instrumentation control** — 7 inspector snapshots with no bingo involved leave the port constant and the target healthy, so the measurement apparatus is not the cause.
- **Zero-instrumentation control** — bingo attach/detach with no inspector at all still wedges the target, so the inspector is not required to reproduce.

*Relationship to #233.* That issue measures ~6 permanently leaked Mach port names per **launched** session. This is a **separate root cause with different severity**: it is on the attach path, its harm is to a third-party process rather than to bingo's own namespace, and it is a hang/kill rather than a leak. The two interact only in the fix: a port-teardown change that destroys `excPort` on a live attached target **before** restoring the prior handler will kill the victim (`ord1` above), so #233's teardown must adopt the ordering in this issue. Neither issue's measurements were used to corroborate the other.

Full raw evidence — helper sources, exact commands, codesign verification, tuple tables, target/handler logs, determinism matrices — is preserved outside the repo in the proof session artifacts (`REPORT.md`, `ADVERSARIAL.md`, `ORDERING.md`, `logs/`); happy to attach any specific transcript on request.

Verdict: **confirmed**, confidence ~0.99.

xsachax:
Fixes #233.

## Dependency and scope

This is the **second layer above #262**, which fixes #234. Its base is `xsachax-darwin-attach-restoration` at verified prerequisite commit `4b6c7c520d5a088dd6a6c8d5117c11c4dfb00205`, not `main`. The user explicitly authorized preparing #233 as a dependent PR after verified #234, overriding only the historical merge-first gate. This layer does not reimplement or weaken the prerequisite's victim-restoration proof. No merge or auto-merge is requested.

## Changes

- Track each acquired Mach right immediately, including partial setup before `portsOK`, cached task rights, separate SEND/RECEIVE ownership, and counted retained thread credits.
- Retire the namespace only after `canReleaseMachNamespace()` confirms durable teardown, actual waiter/diagnostic acknowledgement, target release, no attached ownership, and absence from `outstandingDarwinDetaches`. Incomplete restoration still blocks every namespace operation.
- Cancel/reconcile dead-name notification ownership before releasing its receiver; reclaim returned send-once rights and the extra dead-name credit created when the notification fires. Release only Bingo-owned urefs using the actual live/dead right type, preserving independently held coalesced refs.
- Deterministically release exception/control/notification rights, port sets, retained threads, and cached task rights after launch/Kill, natural exit, and successful attach/detach. Preserve the prerequisite's already-completed exception SEND barrier and prior-handler ownership without double-release.
- Retain partial namespace failures as `ErrBackendCleanupIncomplete`: no failed name is forgotten, the engine admits cleanup-only retries, late waiter results cannot re-enter stop handling, and hub shutdown retains the cleanup owner until completion.
- Document ownership, release order, retry/error behavior, and regression requirements in `AGENTS.md` and `docs/ErrorHandling.md` in the implementation commit.
- Follow up the two CI cognitive-complexity findings with a separate fix commit: extract the loop's terminal retirement gate and flatten the four lifecycle subtests into a table with cohesive event-collection/assertion helpers. Every condition and assertion is preserved; lint thresholds, waiter ordering, cleanup ownership, and the prerequisite are unchanged.

## Exact-head verification

Final tested, committed, and pushed head: **`ed3f7aa116b5ee0111dc76c66905cf8c76f7fbb2`**. This supersedes the earlier validation of `2cbaef9e2caf37578e1461229e2f197d8b01a12f`.

**The reported CI failure is resolved on this head.** Hosted [Go run 35284288761](https://github.com/bingosuite/bingo/actions/runs/35284288761) passed both jobs: build in 36s, unit tests in 1m14s. Its build log confirms the configured **golangci-lint v2.13.2 reports 0 issues**; imports, vet, build, Linux unit tests, and Linux stop-TID race regressions all passed. Locally, the installed v2.11.3 reproduced both original gocognit errors and reports 0 issues after the refactor using the unchanged configuration and Linux-target analysis.

The full native Apple Silicon suite was rebuilt on the final committed head with the trusted Darwin recipe, debugger-entitled codesigned, signature-verified, and actually executed with the race detector: **63 passed, 0 failed, 1 existing opt-in over-cap skip**, 352.142 seconds, seed `1789685252`. This includes all 11 new namespace cases and the prerequisite's restoration cases, 200-iteration churn, 40-stop hygiene, stepping, Pause, Kill, full-stack WebSocket, and DAP coverage. Existing task-send hygiene remained **1 -> 1**.

The final-head namespace census ran all required cycles in one long-lived debugger process (PID `92298`): **20 launches** (10 natural exits with actual status 42, 10 Kills split between running and breakpoint-stopped targets) and **10 real live attach/detaches** (5 custom-handler, 5 null-handler; running and breakpoint-stopped cases). Every captured Bingo-owned name disappeared. Attached targets retained correct prior handlers/original bytes, delivered genuine post-detach traps, made worker/heartbeat progress, and exited normally rather than being killed to make cleanup pass.

| Census point | Port sets | Dead names |
| --- | ---: | ---: |
| Baseline | 0 | 0 |
| Deliberately allocated test-owned leak | 1 | 1 |
| Leak reclaimed, every one of 30 cycles, and final | 0 | 0 |

The deliberate-leak self-check also verifies individual names and idempotent cleanup. Total name count changed from 34 to 22 with runtime thread activity; acceptance uses exact port-set/dead-name counts and owned names, not that noisy total.

Deterministic coverage includes all 11 setup-failure boundaries, release-failure retries, every COMPLETE predicate, SEND-to-DEAD transitions, notification credits, external refs, idempotency, exact-waiter timeout handling, engine lifecycle retention, and hub disposal. Native fault cases cover partial setup/unwind, retained post-COMPLETE release failure while the victim progresses, outstanding RPC ownership blocking destruction, and independently owned task/thread refs surviving cleanup.

Development mutation checks deliberately omitted (separately) production port-set release and the extra notification dead-name credit. **Both failed on the first measured launch with a still-present Bingo-owned name.** Both mutations were removed before committing and final exact-head verification.

```sh
go vet -tags bingonative ./internal/debugger/... ./internal/hub/... ./internal/server/... ./internal/dap/...
go test -tags bingonative -race ./internal/debugger/... ./internal/hub/... ./internal/server/... ./internal/dap/... -count=1 -timeout=180s

env CGO_ENABLED=1 go test -tags 'e2e bingonative' -race -c -o ./build/bingo-e2e.test ./test/integration
codesign --sign - --entitlements entitlements.plist --force ./build/bingo-e2e.test
codesign --verify --strict ./build/bingo-e2e.test
env BINGO_E2E_CHURN_ITERS=200 BINGO_E2E_HYGIENE_ITERS=40 ./build/bingo-e2e.test -test.v -test.timeout 600s

env GOOS=linux GOARCH=amd64 CGO_ENABLED=0 golangci-lint run --config .golangci.yml ./...
env GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go vet ./...
env GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c -o ./build/bingo-debugger-linux.test ./internal/debugger
```

All commands passed on the final committed head. Relevant race package times: debugger 26.152s, hub 7.317s, server 7.714s, DAP 10.906s. The temporary Linux test binary was inspected as x86-64 ELF and removed. Local Linux-target lint used v2.11.3; hosted Linux CI independently passed with v2.13.2 as recorded above.

## Limits

Native Linux execution is unavailable locally; local Linux proof is target-specific lint, cross-vet and test compilation, not native execution. The hosted Go workflow provides Linux build/unit/race execution. The established exclusive debugger/RPC/control/notification ownership premise remains. A deadline cannot interrupt a synchronous Mach call already hung in the kernel. The human-only `darwin-e2e-verified` label has not been applied.


xsachax:
Post-merge synchronization evidence for head `073607e32336297874c178ef25ce39be1bfdfcef`:

- Replayed only the two upper commits onto landed main `16da04639a1a69b9235356ba29fcec9801fbf729`. Both patches, authors and commit messages are unchanged; the entire tracked tree remains `cf09898b76df46e26e6a9f2df692b937b9f096f3`, identical to the previously verified upper head. No source conflict resolutions or source edits were needed.
- Fresh full `just e2e-darwin` passed on this exact published head, 2026-09-18 05:35:34–05:41:18 UTC: 63 passed, zero failures, one existing opt-in skip. Normal defaults, fresh race-enabled compile and signing; 20-launch/10-attach native census and deliberate-leak control passed. Relevant debugger/hub/server/DAP race suites and vet also passed. Main, head, trusted recipe/signing inputs and the clean worktree were revalidated afterward.
- Remaining blocker is GitHub native-stack metadata: this PR still targets the already-merged lower branch at `25374c5e`, rather than main. Both supported REST and GraphQL base updates reject the repair with “Cannot change the base branch because the pull request is part of a stack.” Stack #264 retains its complete [#262, #263] history; no unstack/recreation, lower-branch rewrite, protection change or fabricated verification status was attempted.

This records native proof for `073607e` only, not for any subsequent rewritten head. The native Rebase stack operation is still needed to reconcile the stale base before landing.

xsachax:
### Conflict repair and landing result

The incomplete history/base-retarget transaction has been rolled back to exact head `3f04c861c6040ef051ba2e3e53d52486ff305390`. No source files changed: its entire tree is identical to the separately verified `073607e32336297874c178ef25ce39be1bfdfcef` tree (`cf09898b76df46e26e6a9f2df692b937b9f096f3`). Both supported base-update APIs rejected retargeting this native-stack member, so leaving the onto-main history against its old parent would have preserved an introduced conflict.

GitHub now reports this restored head **MERGEABLE**, with **StackCondition PASSED**. Stack #264 remains exactly [#262 merged, #263 open]; main and the lower branch were not changed. The earlier actual `3f04` native Darwin run passed 63 specs with one existing opt-in skip. The separate `073` run is retained as historical evidence, not relabeled as a run of the current head.

One supported native `merge_queue` request was submitted on this exact restored head, with no custom merge method or bypass parameter. Operation `3dc10a1f-2312-48a7-8542-203ee838ed5f` returned **failed** because the required approving reviews are missing (the one- and two-approval rules). Postcondition reads confirm nothing merged or remained queued, and main is still `16da04639a1a69b9235356ba29fcec9801fbf729`.

The conflict is repaired; landing is not complete. The native REST merge API provides no administrator-bypass option. No protections, reviews, verification labels, or stack membership were changed, and the failed merge mutation will not be retried without a material readiness change.

xsachax:
The approved main-based replacement, #267, has now merged into `main` at `61c0a05b1d1be4fda66decbd9679d1de1bdaf152` (2026-09-19 16:35:33 UTC), using the explicitly authorized administrator rebase after all 19 checks passed.

It carries exactly this PR’s two upper-layer patches. Post-merge Git and API checks confirm the complete landed tree is `cf09898b76df46e26e6a9f2df692b937b9f096f3`, identical to this PR at `3f04c861c6040ef051ba2e3e53d52486ff305390`; #233 is closed. The replacement head also received its own full native Darwin run (63 passed, 0 failed, 1 skipped), rather than inheriting an old-head attestation. The replacement description records the exact-head proof and successful trusted verification workflow.

This was the authorized fallback for GitHub’s refusal to retarget an existing native-stack member. Stack #264, its complete [262, 263] membership, both original branches and this PR’s base/head remain unchanged. #263 has not itself merged or been retargeted; its code changes are now fully landed through #267.

